PR #138, opened 2026-06-29
fix: enforce read-only DB, make shared-DB concurrency-safe, and address review findings

nao1215:
## Summary

Addresses the review findings, one fix per commit (TDD where applicable):

1. **High — `ReadOnlyDB` was not actually read-only.** `Query*`/`QueryRow*` passed SQL through unchecked (so `DELETE ... RETURNING` mutated data), and detection only matched the first keyword (so `/*x*/ DELETE` and `WITH ... DELETE` slipped through). Now the whole statement is scanned for top-level write keywords — skipping comments, string literals, quoted identifiers, and parenthesised CTE bodies — and writes are rejected on the Query paths of both `ReadOnlyDB` and `ReadOnlyTx`.
2. **High — the returned `*sql.DB` was not safe to share across goroutines.** `Open` backed the database with a single bare `:memory:` connection reused for every pooled connection, which raced (and could crash) when shared. It now uses a uniquely-named shared-cache in-memory database (`file:...?mode=memory&cache=shared`) so each pooled connection gets its own real connection to the same data; access is serialised per connection by `database/sql` and nested queries (querying while iterating `*sql.Rows`) still work. The README's segfault warnings are replaced with accurate concurrency guidance in all 7 languages. Auto-save keeps its single-connection design (its save hook fires on connection close), now documented on `EnableAutoSave`.
3. **Medium — `make lint` was red (40 findings).** All `goconst`/`intrange`/`copyloopvar`/`rowserrcheck`/`noctx`/`gofmt`/`staticcheck` issues are fixed; lint is now green. No behavioural change (extracted constants keep their literal values).
4. **Medium — README oversold the implementation.** The tagline no longer claims files are queried "directly without imports"; it explains files are loaded into an in-memory SQLite database. A new *Memory and streaming* note states which formats stream (CSV/TSV/JSON arrays) and which are read in full (LTSV, non-array JSON, Parquet, Excel). Applied to all 7 languages.
5. **Medium — no documentation drift detection.** Adds `doc_sync_test.go` (like gup/sqly): every translation must exist, share language-independent markers, match the English top-level section count, and link every sibling language.
6. **Low — single-package design.** First step toward decomposition: the self-contained SQL write-detection logic is extracted into a focused `internal/sqlguard` package with its own tests. The public API is unchanged; the larger `builder.go` split is left for a dedicated change.

## Testing

- `go test -race ./...` — green
- `make lint` — 0 issues

New tests: read-only write-rejection (incl. comments/CTE/RETURNING), concurrent + nested-query race tests, `internal/sqlguard` table tests, and README drift checks.

## Summary by CodeRabbit

* **New Features**
  * Improved concurrent access support so the database handle can be shared safely across goroutines.
  * Added broader protection against accidental writes in read-only mode.

* **Bug Fixes**
  * Fixed in-memory database handling to avoid cross-instance data sharing within the same process.
  * Improved detection of write queries, including cases hidden in comments or nested statements.

* **Documentation**
  * Updated README content across languages with clearer memory, streaming, and concurrency guidance.
  * Added tests to keep translated READMEs in sync with the main documentation.


nao1215:
@coderabbitai review

nao1215:
Addressed the misspell findings (US spelling: serialize/sanitizes/honoring/parenthesized). While doing so I also caught a doc/impl mismatch: the Concurrency section still described the old `SetMaxOpenConns(1)` pinning approach, but the implementation now uses a shared-cache in-memory database — rewrote that section in all 7 languages to match the code.

nao1215:
Addressed the CodeRabbit findings:

- **sqlguard (security + maintainability)**: `IsWrite` now also blocks SQLite mutators — assigning `PRAGMA` (e.g. `PRAGMA foreign_keys = ON`), `ATTACH`/`DETACH`, `VACUUM`, `ANALYZE`, `REINDEX` — while keeping reading PRAGMAs allowed. Replaced the mutable package-level map with an `isWriteKeyword` switch (no global state).
- **readonly.go QueryRow sentinel**: appended `WHERE 1 = 0` so it always yields no row even if such a table name exists.
- **ReadOnlyStmt query paths**: mirrored the write guard on `Query`/`QueryContext`/`QueryRow`/`QueryRowContext` as defense in depth (note: `PrepareContext` already refuses to prepare any write, so a ReadOnlyStmt can't normally wrap one — added tests confirming DML+RETURNING, CTE writes, and the SQLite mutators are all rejected at Prepare).
- **concurrency_test.go**: timeout context for fast-fail on deadlock regressions, `db.Close()` errors handled, stale single-connection comment updated.
- **doc_sync_test.go**: documentation file paths built with `filepath.Join`.
- **Taglines**: added JSON/JSONL — the gap existed in all six translations (not just es/fr), fixed all.

`go test -race ./...` and `make lint` are green.

Issue #28, opened 2025-08-31
Version 1.0.0 Roadmap (Task List)


nao1215:
This roadmap outlines the strategic development direction for filesq;, focusing on community growth, stability, and enhanced usability. I will try to receive bug reports and contributions from many people.

# Primary Goals
## Goal 1: Achieve 1,000 GitHub Stars
This is a [personal bucket list](https://github.com/nao1215/my-life?tab=readme-ov-file#bucket-list-%E6%AD%BB%E3%81%AC%E3%81%BE%E3%81%A7%E3%81%AB%E3%81%97%E3%81%9F%E3%81%84%E3%81%93%E3%81%A8) goal, and among all my open-source projects, the filesql has the highest potential to reach this milestone. The second most promising project is [nao1215/gup](https://github.com/nao1215/gup).  However, since I have low motivation to continue developing gup, I will prioritize filesql.

## Goal 2: Achieve Stable Operations
Focus on reducing bugs and ensuring reliable operation in CLI and one-shot batch processing scenarios.

## Goal 3: Clarify Use Cases and Expand Documentation
Provide comprehensive documentation with real-world examples to help users understand when and how to use filesql effectively.

# Roadmap (Task List)

When something comes to mind, I add it as a task. Suggestions are welcome :)

## Development
- [x] Create scenario-based test cases aligned with common use cases
  - #42
- [x] Adopt testify/assert to improve test code quality and maintainability
  - #33 
- [x] Expand supported file extensions
  - [x] Add support for custom delimiters in CSV/TSV files per input source
- [x] Bug fixes and performance optimization for CLI operations
  - #39 
  - #38
  - #37
  - #36
- [x] In the present day, the mainstream development style is to use LLMs, with configuration files for each tool (e.g., Claude Code) added to the project.
  - #29 
- [x] As a sample for filesql, I will integrate filesql into [sqly](https://github.com/nao1215/sqly) and [sqluv](https://github.com/nao1215/sqluv). There is also a possibility of merging sqly and sqluv to create a new command.
   - https://github.com/nao1215/sqly/pull/142
   -  sqluv will be public archive

## Community & Documentation
- [ ] Improve CONTRIBUTING.md with detailed guidelines
- [ ] Create "good first issue" labels and beginner-friendly tasks
- [ ] Develop a landing page

## Promotion
- [x] [Reddit](https://www.reddit.com/r/golang/comments/1mymsl1/filesql_a_go_sql_driver_for_csvtsvltsv_files/)
[...]

PR #209, opened 2026-08-01
refactor: FileType names a format, not a format-and-codec combination

### LINKED ISSUE #208: FileType fuses format and compression: 56 of its 65 constants name a combination
## Problem

`FileType` fuses two independent things: the input format and the codec wrapping it. That produces the cross product, so **56 of its 65 constants name a combination rather than a format**:

```
FileTypeCSVGZ, FileTypeCSVBZ2, FileTypeCSVXZ, FileTypeCSVZSTD, FileTypeCSVZLIB,
FileTypeCSVSNAPPY, FileTypeCSVS2, FileTypeCSVLZ4,   ... and the same eight for
TSV, LTSV, Parquet, XLSX, JSON, and JSONL
```

They are the single largest part of this package's public API: 314 exported symbols in the root package, and these are 56 of them. Every switch over the type has to list them, which is why `stream.go`'s `createDecompressedReader` is 70 lines of eight arms naming seven formats each — a second implementation of what `CompressionHandler.CreateReader` already does from a `CompressionType`.

The naming is also ambiguous where the two axes collide: `FileTypeJSONLZ4` is lz4-compressed **JSON**, while `FileTypeJSONLLZ4` is lz4-compressed **JSONL**. One `L` apart, and nothing in the name says which reading is right.

## Why they exist

Only one caller needs them: `AddReader`.

```go
func (b *DBBuilder) AddReader(reader io.Reader, tableName string, fileType FileType) *DBBuilder
```

An `io.Reader` carries no path, so nothing can be inferred from an extension. `FileTypeCSVGZ` is the only way a caller can say "these bytes are a gzipped CSV".

Every other path already ignores them. `stream_processor.go` passes the base format and a reader it has already decompressed, so `createDecompressedReader` falls through to its no-op default for files. `parser.BaseFileType` exists precisely to fold a fused constant back to its format.

## Design

Give `AddReader` functional options, which is non-breaking — the existing three-argument call keeps compiling, so no README example or user code changes.

```go
// ReaderOption configures one reader added with AddReader.
//
// It is the place per-source settings go. A reader has no path, so anything a
// file's name would have answered — which codec wraps it, and in time which
// delimiter separates its fields — has to be stated, and stating it here keeps
// AddReader's ordinary three-argument call unchanged.
type ReaderOption func(*readerInput)

// WithCompression declares the codec wrapping a reader passed to AddReader.
func WithCompression(compression CompressionType) ReaderOption

func (b *DBBuilder) AddReader(reader io.Reader, tableName string, fileType FileType, opts ...ReaderOption) *DBBuilder
```

Call site:

```go
gz, _ := os.Open("users.csv.gz")
builder.AddReader(gz, "users", filesql.FileTypeCSV, filesql.WithCompression(filesql.CompressionGZ))
```

`FileType` then names formats only:

```go
const (
	FileTypeCSV FileType = iota
	FileTypeTSV
	FileTypeLTSV
	FileTypeParquet
	FileTypeXLSX
	FileTypeJSON
	FileTypeJSONL
	FileTypeACH
	FileTypeFedWire
	FileTypeUnsupported
)
```

The option pattern is chosen over a fourth positional parameter because it is non-breaking, and over an `AddReaderWithCompression` sibling because the next per-source setting — a custom delimiter, already on the v1.0.0 roadmap (#28) — then has somewhere to go without another method or another signature change.

## Expected end state

- `FileType` has **10 constants instead of 65**.
- `stream.go`'s `createDecompressedReader` is one line: `return NewCompressionHandler(p.compression).CreateReader(reader)`.
- `parser_bridge.go`'s `filesqlToParserFileTypes` has 7 entries instead of 65, and `filesqlFileType` folds compression away with `parser.BaseFileType`.
- `FileType.extension()` returns the format's extension only; the codec's suffix comes from `CompressionType.Extension()`.
- `FileType.baseType()` is gone — every `FileType` is a base type — along with the deprecated `getBaseFileType` wrapper.
- `file.toTable()` reads through `f.openReader()` (which unwraps the codec) instead of opening the raw file, because the parser is now handed bytes that are already the format.
- `AddReader(r, name, FileTypeCSV)` still compiles and behaves as before.

## Verified so far

An attempt got the production code to this state: it builds, `go vet` is clean, and all seven non-root packages pass (`dialect`, `frame`, `internal/sqlguard`, `parser`, `parser/ach`, `parser/wire`, `prep`). It was reverted rather than landed half-migrated, because the test suite needs real work rather than a rename.

## Remaining work

The 259 references in `*_test.go` are not all mechanical:

1. **`TestParserFileType` needs redesigning, not renaming.** It asserts the fused map entry by entry; with the map holding formats only, the property worth testing is that `filesqlFileType(parser.CSVGZ) == FileTypeCSV` — compression folded away — plus a round trip for the seven formats.
2. **`TestFileType_Extension`, `TestGetFileExtension`, `TestFileTypeString`**: drop the rows whose expected value carries a codec (`".csv.gz"`, `"CSV (gzip)"`). Those asserted the fused enum.
3. **The compression tests must carry their codec per row.** `TestCreateDecompressedReader_AllCompressionTypes`, `TestCreateDecompressedReader_TSVCompressionTypes`, `TestCreateDecompressedReader_LTSVCompressionTypes`, `TestStreamingParser_ParseFromReader_CompressedJSON`, `TestStreamingParser_ProcessInChunks_CompressedJSON`, and `TestCreateDecompressedReader_InvalidData` each build compressed bytes and previously named the codec through the `FileType`. Their tables need a `compression CompressionType` field threaded into `newStreamingParser`. A blanket `CompressionNone` compiles and then fails at parse time with "invalid character '\x1f'", so this is the step to do carefully rather than by regex.
4. **One contract change to pin deliberately.** `createDecompressedReader` for `CompressionNone` now returns a non-nil no-op close function, because that is `CompressionHandler.CreateReader`'s convention, where the old switch returned `nil`. `TestCreateDecompressedReader/unsupported_compression` asserts the old one.
5. **`prep/alias.go`** re-exports 56 of `parser`'s compressed constants. It aliases the `parser` enum, not this one, so it can be handled independently — but leaving it means the module still exposes the cross product under a second name.
6. Docs: a `WithCompression` example in the README's reader section, and a note that `FileType` names formats only.

## Not in scope

`parser.FileType` keeps its fused constants. That package is the lower level where a "format-and-codec" name is the honest representation of what it dispatches on, and `parser.BaseFileType` already exists to fold one back.

nao1215:
Closes #208.

## What

`FileType` fused two independent things: the input format and the codec wrapping it. That produced the cross product, so **56 of its 65 constants named a combination rather than a format**. They were the single largest part of this package's public API — 314 exported symbols in the root package, and these were 56 of them.

The naming was also ambiguous where the two axes collided: `FileTypeJSONLZ4` was lz4-compressed **JSON**, while `FileTypeJSONLLZ4` was lz4-compressed **JSONL**. One `L` apart, and nothing in the name said which reading was right.

Only one caller needed them: `AddReader`, because an `io.Reader` carries no path and nothing can be inferred from an extension. Every other path already ignored them — `stream_processor.go` passes the base format and a reader it has already decompressed, and `parser.BaseFileType` exists precisely to fold a fused constant back to its format.

## Design

`AddReader` takes functional options, which is non-breaking:

```go
gz, _ := os.Open("users.csv.gz")
builder.AddReader(gz, "users", filesql.FileTypeCSV, filesql.WithCompression(filesql.CompressionGZ))
```

The existing three-argument call keeps compiling and behaving as before, so no README example or user code reading uncompressed bytes changes. The option pattern was chosen over a fourth positional parameter because it is non-breaking, and over an `AddReaderWithCompression` sibling because the next per-source setting — a custom delimiter, on the v1.0.0 roadmap (#28) — then has somewhere to go without another method.

## Result

- `FileType` has **10 constants instead of 65**.
- `stream.go`'s `createDecompressedReader` is one line: `return NewCompressionHandler(p.compression).CreateReader(reader)`. It was 70 lines of eight arms naming seven formats each — a second implementation of what `CompressionHandler.CreateReader` already did.
- `parser_bridge.go`'s `filesqlToParserFileTypes` has 7 entries instead of 65, and `filesqlFileType` folds compression away with `parser.BaseFileType`.
- `FileType.extension()` returns the format's extension only; the codec's suffix comes from `CompressionType.Extension()`.
- `FileType.baseType()` and the deprecated `getBaseFileType` wrapper are gone — every `FileType` is a base type.
- `file.toTable()` reads through `f.openReader()`, because the parser is now handed bytes that are already the format.

One contract change is deliberate: `createDecompressedReader` for an uncompressed source returns a **non-nil no-op** close function, because that is `CompressionHandler.CreateReader`'s convention. The old switch returned `nil`. `TestCreateDecompressedReader` now pins the new behavior and says why.

## Tests

The 259 references in `*_test.go` were not mechanical.

- **`TestParserFileType` was redesigned, not renamed.** It asserted the fused map entry by entry. It is now a round trip over the seven formats plus `TestFilesqlFileType_FoldsCompression`, a table of all 56 compressed parser constants checking that each folds to its format — the property the map used to state one row at a time.
- **The compression tables carry their codec per row.** `TestCreateDecompressedReader_{All,TSV,LTSV}CompressionTypes`, the JSON/JSONL compressed tests, and `TestCreateDecompressedReader_InvalidData` each build compressed bytes and previously named the codec through the `FileType`. A blanket `CompressionNone` compiles and then fails at parse time with `invalid character '\x1f'`, so each row was threaded individually.
- **Those tables now assert the header, not just the rows.** Verified by setting every row to `CompressionNone`: with the old assertions, the snappy and s2 rows still **passed**, because both leave short incompressible input nearly verbatim after their frame magic — only the header carried the corruption. With the header asserted, all 15 compressed rows fail under that mutation and only the genuinely-uncompressed row passes.
- **`TestFileTypeExtension_WithCompression` is new**, walking the full 7x9 grid: `format.extension() + codec.Extension()` gives the compound suffix, and that path detects back to the same pair.
- **The XLSX and Parquet "compressed" tests were fiction.** Both fed the parser uncompressed bytes under a compressed `FileType` and logged whatever happened, so neither established that a compressed workbook or Parquet file loads at all. They now really compress and read back, asserting header and records.
- **`TestDBBuilder_AddFS_Compressed` is new.** `AddFS` hands a file over still wrapped, so the codec has to travel alongside the format on the reader input — and nothing covered that. Verified it fails when the `file_processor.go` wiring is removed.

## Scope

`parser.FileType` keeps its fused constants. That package is the lower level where a format-and-codec name is the honest representation of what it dispatches on, and `parser.BaseFileType` already exists to fold one back.

`prep/alias.go` re-exported 56 of those under a second spelling, which meant the module still exposed the cross product twice. It now re-exports formats only; nothing in the repo referenced the compressed aliases.

## Verification

- `go test ./...` green across all 8 packages; `go vet ./...` clean; `golangci-lint run ./...` reports 0 issues.
- sqly built and its full suite run against this branch via `go mod edit -replace` — it does not reference `filesql.FileType` at all, so nothing there changes. The go.mod edit was not committed.

## Summary by CodeRabbit

* **New Features**
  * Added optional compression settings for reader imports, supporting gzip, xz, zstd, zlib, snappy, s2, and lz4.
  * Added compressed-file loading through filesystem imports.
  * Added an example for loading gzip-compressed data.

* **Improvements**
  * File format and compression are now configured separately.
  * Compressed and uncompressed inputs are handled consistently, including streaming and chunked reads.

* **Breaking Changes**
  * Removed compression-specific file type constants and aliases. Use the base file format with a compression option.

* **Documentation**
  * Updated the changelog and README with the new configuration workflow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- `FileType` names a format and no longer fuses in the codec wrapping it, so it has 10 constants instead of 65. 56 of the old ones named a combination — `FileTypeCSVGZ`, `FileTypeCSVBZ2`, and the same eight codecs for TSV, LTSV, Parquet, XLSX, JSON, and JSONL — which made them the single largest part of this package's public API, and the naming was ambiguous where the two axes collided: `FileTypeJSONLZ4` was lz4-compressed JSON while `FileTypeJSONLLZ4` was lz4-compressed JSONL, one `L` apart with nothing in the name to say which reading was right. `AddReader` was the only caller that needed them, because an `io.Reader` carries no path to infer a codec from; it now takes options, and the codec is stated with `WithCompression(CompressionGZ)`. The existing three-argument call is unchanged, so a caller reading uncompressed bytes needs no edit. `prep` re-exported 56 of the parser's compressed constants under a second spelling and now re-exports formats only; `parser.FileType` keeps its fused constants, since that is the lower level where a format-and-codec name is the honest description of what it dispatches on, and `parser.BaseFileType` folds one back.
+- `stream.go`'s `createDecompressedReader` was 70 lines of eight arms naming seven formats each — a second implementation of what `CompressionHandler.CreateReader` already did from a `CompressionType` — and is now one line that calls it. One behavior follows from that: for an uncompressed source it returns a no-op close function rather than `nil`, which is `CreateReader`'s convention, so a caller can invoke it unconditionally.
+
 ## [0.29.0] - 2026-07-30
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -60,7 +60,12 @@ codec that is. A reader has no path, so `AddReader` takes the codec as an
 option:
 
 ```go
-gz, _ := os.Open("users.csv.gz")
+gz, err := os.Open("users.csv.gz")
+if err != nil {
+	return err
+}
+defer gz.Close()
+
 builder.AddReader(gz, "users", filesql.FileTypeCSV, filesql.WithCompression(filesql.CompressionGZ))
 ```
 
--- a/README.md
+++ b/README.md
@@ -54,6 +54,19 @@ Compressed wrappers are supported for CSV, TSV, LTSV, JSON, JSONL, Parquet, and 
 
 ACH and Fedwire do not use external compression wrappers.
 
+Format and compression are separate. `FileType` names the format only —
+`FileTypeCSV` is a CSV whether or not a codec wraps it — and a path says which
+codec that is. A reader has no path, so `AddReader` takes the codec as an
+option:
+
+```go
+gz, _ := os.Open("users.csv.gz")
+builder.AddReader(gz, "users", filesql.FileTypeCSV, filesql.WithCompression(filesql.CompressionGZ))
+```
+
+Without `WithCompression` the reader's bytes are read as the format directly,
+so the ordinary three-argument call is unchanged.
+
 ## Installation
 
 ```bash
@@ -378,6 +391,7 @@ The GoDoc examples are fully tested with `go test`. The tables below show the fa
 | Open files and query them | `ExampleOpen`, `ExampleOpenContext` | [example_api_test.go](./example_api_test.go), [example_test.go](./example_test.go) |
 | Load files into an existing `*sql.DB` | `ExampleLoadInto`, `ExampleDBBuilder_LoadInto` | [example_api_test.go](./example_api_test.go) |
 | Build from readers, paths, or embedded FS | `ExampleNewBuilder`, `ExampleDBBuilder_AddReader`, `ExampleDBBuilder_AddPath`, `ExampleDBBuilder_AddFS` | [example_test.go](./example_test.go) |
+| Read a compressed reader | `ExampleDBBuilder_AddReader_compressed` | [example_test.go](./example_test.go) |
 | Tune chunked loading | `ExampleDBBuilder_SetDefaultChunkSize` | [example_api_test.go](./example_api_test.go) |
 | Handle malformed CSV/TSV rows | `ExampleDBBuilder_WithMalformedRowPolicy` | [example_api_test.go](./example_api_test.go) |
 | Attach your own logger | `ExampleDBBuilder_WithLogger`, `ExampleNewSlogAdapter` | [example_api_test.go](./example_api_test.go) |
--- a/builder.go
+++ b/builder.go
@@ -79,9 +79,31 @@ type readerInput struct {
 	tableName string
 	// fileType specifies the file format using domain/model types
 	fileType FileType
+	// compression is the codec wrapping the reader, CompressionNone when the
+	// bytes are already the format.
+	compression CompressionType
 	// closer is an optional closer for the reader (set when the reader was opened internally, e.g. from AddFS).
 	// User-provided readers (from AddReader) do not set this field.
 	closer io.Closer
+}
+
+// ReaderOption configures one reader added with AddReader.
+//
+// It is the place per-source settings go. A reader has no path, so anything a
+// file's name would have answered — which codec wraps it — has to be stated,
+// and stating it here keeps AddReader's ordinary three-argument call unchanged.
+type ReaderOption func(*readerInput)
+
+// WithCompression declares the codec wrapping a reader passed to AddReader.
+//
+// Example:
+//
+//	gz, _ := os.Open("users.csv.gz")
+//	builder.AddReader(gz, "users", filesql.FileTypeCSV, filesql.WithCompression(filesql.CompressionGZ))
+func WithCompression(compression CompressionType) ReaderOption {
+	return func(input *readerInput) {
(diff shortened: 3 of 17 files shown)


nao1215:
Both findings checked.

**`file_processor.go` — header-only compressed FS reader.** Not reproducible. `processDelimitedInChunks` has an explicit header-only branch: when the reader yields a header and no records, it emits a chunk carrying the header with `records: nil`, so `createTableFromChunk` creates the table and `createEmptyTable` is never reached. The second decompressor over a consumed stream is therefore not constructed on this path.

Worth noting the behaviour is also unchanged by this PR: before it, `AddFS` set `fileType` to the fused `FileTypeCSVGZ` and `createDecompressedReader` built the gzip reader from that; now it sets `FileTypeCSV` plus `CompressionGZ` and builds the same reader from the pair. Same reader, same number of times.

That said, nothing covered the case, so it is pinned now — `TestDBBuilder_AddFS_CompressedHeaderOnly` loads a header-and-no-rows file through `AddFS` under none/gz/zstd/lz4 and asserts the table exists with the header's columns and zero rows.

**`README.md` — the opened file.** Fair, and applies to the godoc on `WithCompression` too. A reader handed to `AddReader` is not closed by the builder, so a snippet worth copying has to check `os.Open` and close it. Both now do.

Fixed in 9c9a9a4.

PR #218, opened 2026-08-02
fix: make every sentinel an error names reachable with errors.Is

### LINKED ISSUE #216: errors.Is only reaches one sentinel: 107 of 108 wraps flatten the inner error with %s
## Problem

A caller cannot use `errors.Is` to find out why a save failed, because the sentinel that names the reason is flattened into text before it leaves the package.

Overwriting a `.tsv.bz2` source in place fails, correctly — bzip2 is read but has no writer here. The message says exactly that:

```
auto-save failed: filesql: I/O operation failed: failed to overwrite /tmp/x/products.tsv.bz2:
filesql: compression operation failed: failed to create writer:
filesql: unsupported file format: bzip2 compression is not supported for writing
```

But:

```go
errors.Is(err, ErrUnsupportedFormat) // false
errors.Is(err, ErrCompression)       // true
errors.Is(err, ErrIOOperation)       // true
```

`ErrUnsupportedFormat` appears in the message and not in the chain, so the one distinction a caller would act on — "this file's format or codec cannot be written, pick another destination" versus "the compressor failed" — is only available by matching on the string.

## Cause

The wrapping loses it one frame in:

```go
// filesql.go
return fmt.Errorf("%w: failed to create writer: %s", ErrCompression, err.Error())
```

`%w` wraps `ErrCompression`; `%s` on `err.Error()` renders the inner error, which is the one carrying `ErrUnsupportedFormat`, as plain text. Everything below that frame is unreachable to `errors.Is` and `errors.As`.

## Scale

This is the house style, not a slip at one site: **107 of the 108 places** that interpolate an inner error do it as `%s` with `err.Error()`.

```
$ grep -rn 'fmt.Errorf("%w: [^"]*: %s"' --include=*.go . | grep -v _test.go | grep -c "err.Error()"
107
```

So each error carries exactly one sentinel — whichever frame is outermost in that call — and every sentinel below it is text. Which one survives depends on where in the call stack the failure was re-wrapped, which is not something a caller can reason about.

## Why it is worth deciding before v1.0.0

The package documents its sentinels as the way to inspect a failure, and the recent API inventory leaned on that when removing `ErrorContext` — the sentinels were named as what replaces it. If `errors.Is` only answers for one arbitrary layer, that promise is weaker than it reads.

It is also the kind of thing that is cheap to change now and awkward later: once callers match on message text because `errors.Is` did not work, changing a message becomes breaking.

## Options

1. **Wrap with `%w` throughout.** `fmt.Errorf("%w: failed to create writer: %w", ErrCompression, err)` — Go 1.20+ allows several. Every sentinel in the stack stays reachable. Mechanical across 107 sites, and it changes messages slightly where `err.Error()` was already prefixed.
2. **Wrap only the inner error and let the sentinel come from below**, so a given failure carries the sentinel nearest its cause. Fewer edits, but a caller then has to know which sentinel each cause uses.
3. **Leave it and document that exactly one sentinel is reachable**, saying which. Honest, and costs a caller the distinction above.

(1) is what I would pick, but it touches the whole package's error paths, so it wants a decision rather than a drive-by.

## Test

`TestAutoSaveOverwriteRefusesCodecItCannotWrite` asserts the present behaviour, including `assert.NotErrorIs(t, err, ErrUnsupportedFormat)`, so whichever way this goes, the change is deliberate and visible in the diff.

Refs #28.

nao1215:
Closes #216.

## The problem

An error could say `filesql: unsupported file format: bzip2 compression is not supported for writing` and still fail `errors.Is(err, ErrUnsupportedFormat)`.

The cause is one frame up:

```go
return fmt.Errorf("%w: failed to create writer: %s", ErrCompression, err.Error())
```

`%w` carries `ErrCompression`. `%s` on `err.Error()` renders everything below it — including the frame carrying `ErrUnsupportedFormat` — as plain text. **115 call sites** did this, so each error carried exactly one reachable sentinel, and which one depended on which frame happened to be outermost. A caller wanting to distinguish "this codec cannot be written, pick another destination" from "the compressor failed" had to match on the message.

That mattered more after #211 removed `ErrorContext`, since the sentinels were named there as what replaces it.

## The fix

Those sites wrap with `%w` instead. Go 1.20+ allows several in one `Errorf`, so the sentinel and the cause both stay in the chain.

**No message changes.** `fmt` renders `%w` through the error's own `Error` method, so `"%w: x: %w"` and `"%w: x: %s"` with `err.Error()` produce the same string:

```
s: "outer: failed: inner: detail"
w: "outer: failed: inner: detail"
same message: true
s reaches inner: false | w reaches inner: true
```

The suite is the evidence: after rewriting all 115 sites, **the entire test suite passed unaltered except one assertion** — `assert.NotErrorIs(t, err, ErrUnsupportedFormat)` in `TestAutoSaveOverwriteRefusesCodecItCannotWrite`, added in #217 precisely so this could not change quietly. If any message had shifted, the tests that match on message text would have failed too.

That assertion becomes:

```go
assert.ErrorIs(t, err, ErrUnsupportedFormat)
assert.ErrorIs(t, err, ErrCompression)
assert.ErrorIs(t, err, ErrIOOperation)
```

— the three frames the failure actually passed through, all reachable now.

## How the rewrite was done

Scripted, with the transformation refusing anything it could not prove safe: it only rewrote a `fmt.Errorf` whose **last argument** is `<expr>.Error()` and whose format string's **last verb** is `%s`, and it reported skips. 115 rewritten, 0 skipped, so no site needed a judgement call.

(`%w` with a nil error renders `%!w(<nil>)`, which would be a real hazard — but every one of these sites is inside an error path where the value is non-nil by construction, and any slip would have shown up as a changed message in the suite.)

## Tests

- **`TestErrorSentinelsAreReachable`** pins the property, not the one path: three failures — a codec with no writer, data that is not the codec it was declared to be, a file this package does not read — each asserted against every sentinel its message names.
- **`TestErrorWrappingKeepsTheMessage`** pins the equivalence the whole conversion rests on, so if it ever stopped holding the reason would be visible.

Verified both are load-bearing by reverting a **single** call site back to `%s`: the suite fails.

## Verification

- `go test ./...` green across all 8 packages; `go vet ./...` clean; `golangci-lint run ./...` reports 0 issues.
- sqly built and its full suite run against this branch via `go mod edit -replace`. The go.mod edit was not committed.

Refs #28.


nao1215:
Also verified end to end: sqly built against this branch with `go mod edit -replace` and run through its full atago suite — **645 + 11 scenarios, 0 failed**. That exercises the changed error paths through a real binary (compression round-trips, the bzip2 write rejection, export inference, write-back), not just filesql's own tests.

PR #491, opened 2026-08-24
fix(loader): load every input under one scope a failure undoes

### LINKED ISSUE #489: A multi-sheet XLSX failing on a later sheet leaves earlier sheets committed and the replaced table lost
Found at f6496a1 by differential testing of load-failure states (S5 streaming/atomicity, filesql-dig-bug session 2026-08-24).

## Phenomenon

`LoadInto` documents that each input is loaded atomically, but a multi-sheet XLSX workbook is loaded sheet by sheet with one committed transaction per sheet, so when a later sheet fails (for example a duplicate header that the reader refuses), the tables of the earlier sheets are already committed to the caller's database, and any same-named table the caller had (which `LoadInto` drops before recreating) is gone for good. The caller receives an error, yet their database has been changed: new `book_good`-style tables appeared, and a pre-existing table with the same name as a loaded sheet has been silently replaced by the workbook's data. This is silent data loss on an API whose documentation promises the opposite. ACH and Fedwire inputs have the same shape of problem: `streamACHFileToDatabase` and `streamWireFileToDatabase` create several tables directly on the executor with no transaction at all, so a failure partway leaves the earlier tables of that file committed.

## Reproduction

```go
package filesql_test

import (
	"context"
	"database/sql"
	"path/filepath"
	"testing"

	"github.com/nao1215/filesql"
	"github.com/xuri/excelize/v2"
	_ "modernc.org/sqlite"
)

func TestXLSXPartialFileFailureLeavesDatabaseAsItWas(t *testing.T) {
	db, err := sql.Open("sqlite", ":memory:")
	if err != nil {
		t.Fatal(err)
	}
	defer db.Close()
	db.SetMaxOpenConns(1)
	if _, err := db.Exec(`CREATE TABLE book_good (v)`); err != nil {
		t.Fatal(err)
	}
	if _, err := db.Exec(`INSERT INTO book_good VALUES ('precious')`); err != nil {
		t.Fatal(err)
	}

	f := excelize.NewFile()
	if err := f.SetSheetName("Sheet1", "good"); err != nil {
		t.Fatal(err)
	}
	if err := f.SetSheetRow("good", "A1", &[]any{"a", "b"}); err != nil {
		t.Fatal(err)
	}
	if err := f.SetSheetRow("good", "A2", &[]any{1, "x"}); err != nil {
		t.Fatal(err)
	}
	if _, err := f.NewSheet("bad"); err != nil {
		t.Fatal(err)
	}
	if err := f.SetSheetRow("bad", "A1", &[]any{"id", "id"}); err != nil { // duplicate header: this sheet is refused
		t.Fatal(err)
	}
	p := filepath.Join(t.TempDir(), "book.xlsx")
	if err := f.SaveAs(p); err != nil {
		t.Fatal(err)
	}

	b, err := filesql.NewBuilder().AddPath(p).Build(context.Background())
	if err != nil {
		t.Fatal(err)
	}
	if err := b.LoadInto(context.Background(), db); err == nil {
		t.Fatal("want error from the bad sheet")
	}

	var kept string
	if err := db.QueryRow(`SELECT v FROM book_good`).Scan(&kept); err != nil {
		t.Errorf("caller's book_good was not left as it was: %v", err)
	}
}
```

Actual output at f6496a1:

```
LoadInto error: /tmp/.../book.xlsx: sheet bad: filesql: duplicate column name: "id" (column 2)
caller's book_good was not left as it was: SQL logic error: no such column: v (1)
```

The load reported an error, but the caller's `book_good` table (schema `(v)`, one row `'precious'`) has been replaced by the workbook's `good` sheet (schema `(a, b)`), and without a pre-existing table the sheet's table is simply left behind in the caller's database after the error.

## Expected behavior

The `LoadInto` documentation in builder.go states: "Each input is loaded atomically: a file that fails partway leaves the database as it was, including the table it was replacing." A workbook is one input, so a workbook whose later sheet fails should leave none of its sheets loaded and should leave any table it was about to replace exactly as it was. The same applies to ACH and Fedwire inputs, which map one file onto several tables.

## Cause (guess)

This is a guess from reading stream_processor.go. `streamXLSXFileToDatabase` calls `loadTable` once per sheet, and `loadTable`'s `beginLoad` opens and commits a transaction per call when the executor is a `*sql.DB`, so each sheet commits on its own and there is no transaction spanning the file; `dropIfReplacing` runs inside the first successful sheet's transaction, which is why the caller's replaced table is gone even though a later sheet failed. `streamACHFileToDatabase` (ach.go) and `streamWireFileToDatabase` (wire.go) never open a transaction and execute DROP/CREATE/INSERT directly on the executor. A fix that covers both `LoadInto` and `LoadIntoTx` is to give every input one atomicity scope in `streamFileToDatabase`/`streamReaderToDatabase`: begin one transaction per input when the executor is a `*sql.DB`, and open a `SAVEPOINT` (released on success, rolled back to on failure) when the executor is a caller-owned `*sql.Tx`, then run all of that input's tables inside it. That also subsumes the staging-table cleanup special case in `abandonLoad`.

## Tests to add

- The reproduction above as a regression test in load_into_test.go: a two-sheet workbook whose second sheet has a duplicate header, loaded with `LoadInto` into a database that already holds a table named like the first sheet's table; assert the error, assert the pre-existing table's schema and rows are untouched, and assert no table named after any sheet exists.
- The sibling without a pre-existing table, same file: after the failed `LoadInto`, assert the database's table list is exactly what it was before the call (no `book_good` left behind).
- The `LoadIntoTx` variant of the same workbook: after the failed call, assert inside the still-open transaction that the caller's tables are as they were, so the caller can continue using the transaction.
- An invariant-style sibling for single-table inputs stays green: a CSV that fails partway through `LoadInto` already leaves the database as it was (this holds today through `loadTable`'s own transaction) and must keep holding once the per-input scope is introduced.

### LINKED ISSUE #490: A file that fails partway through LoadIntoTx leaves a partial table in place of the caller's table
Found at f6496a1 by differential testing of load-failure states (S5 streaming/atomicity, filesql-dig-bug session 2026-08-24).

## Phenomenon

When a file path input fails partway through `LoadIntoTx`, the caller's still-open transaction is left holding a partially loaded table under the input's name in place of the caller's own same-named table, which was dropped by the replacement step before the failure. A reader input behaves differently and better: it is staged, so the same failure leaves the caller's table untouched and `abandonLoad` drops the staging table. So the state a caller finds in their transaction after an error depends on whether the input was a path or a reader, and in the path case it is the worst possible one: their table is gone and an impostor with the file's schema and a truncated row set stands in its name. A caller who catches the error and commits the rest of their transaction commits silent data corruption.

## Reproduction

```go
package filesql_test

import (
	"context"
	"database/sql"
	"fmt"
	"os"
	"path/filepath"
	"strings"
	"testing"

	"github.com/nao1215/filesql"
	_ "modernc.org/sqlite"
)

func TestLoadIntoTxFileFailureLeavesCallersTable(t *testing.T) {
	db, err := sql.Open("sqlite", ":memory:")
	if err != nil {
		t.Fatal(err)
	}
	defer db.Close()
	db.SetMaxOpenConns(1)
	if _, err := db.Exec(`CREATE TABLE tbl (old)`); err != nil {
		t.Fatal(err)
	}
	if _, err := db.Exec(`INSERT INTO tbl VALUES ('keep-me')`); err != nil {
		t.Fatal(err)
	}

	var sb strings.Builder
	sb.WriteString("a,b\n")
	for i := range 3000 {
		fmt.Fprintf(&sb, "%d,value-%d\n", i, i)
	}
	sb.WriteString("badrow-only-one-field\n") // refused by the default MalformedRowStop policy
	p := filepath.Join(t.TempDir(), "tbl.csv")
	if err := os.WriteFile(p, []byte(sb.String()), 0600); err != nil {
		t.Fatal(err)
	}

	b, err := filesql.NewBuilder().AddPath(p).SetDefaultChunkSize(100).Build(context.Background())
	if err != nil {
		t.Fatal(err)
	}
	tx, err := db.Begin()
	if err != nil {
		t.Fatal(err)
	}
	defer tx.Rollback()
	if err := b.LoadIntoTx(context.Background(), tx); err == nil {
		t.Fatal("want error")
	}

	var kept string
	if err := tx.QueryRow(`SELECT old FROM tbl LIMIT 1`).Scan(&kept); err != nil {
		t.Errorf("caller's tbl is not as it was inside the transaction: %v", err)
	}
}
```

Actual output at f6496a1:

```
LoadIntoTx error: /tmp/.../tbl.csv: filesql: column count mismatch: row 3001 has 1 fields, want 2
tables inside tx after failure: [existing tbl]
tbl row count inside tx: 3000
tbl no longer has caller's schema: SQL logic error: no such column: old (1)
```

The caller's `tbl` (schema `(old)`, one row `'keep-me'`) is gone; a table named `tbl` with the file's schema `(a, b)` and the 3000 rows read before the bad row stands in its place inside the caller's transaction. The same scenario through `AddReader` instead of `AddPath` leaves the caller's `tbl` intact.

## Expected behavior

A failed input should leave the caller's transaction as that input found it, whichever kind of input it was. `abandonLoad`'s own comment states the intent for the caller-owned-transaction case: the staging table "is dropped from it rather than left for the caller to find" -- the typed path should hold itself to the same standard, and the caller's replaced table must not be lost while the load can still fail. `LoadInto` already documents per-input atomicity ("a file that fails partway leaves the database as it was, including the table it was replacing"), and `LoadIntoTx` differing from `LoadInto` on what a failed input leaves behind is documented nowhere.

## Cause (guess)

This is a guess from reading stream_processor.go. Path inputs can be re-read, so they go through `loadTyped`, which calls `dropIfReplacing` (dropping the caller's table) and then creates and fills the real table as chunks arrive; when the executor is a caller-owned `*sql.Tx`, `beginLoad` returns it with `ownTx=false` and `abandonLoad` cannot roll back, so it only drops the staging table -- which the typed path never made -- and the partial real table plus the earlier drop of the caller's table survive in the transaction. Reader inputs go through `loadStaged`, where both `dropIfReplacing` and the copy into the real table happen in `declareTable` after the whole input has been read, so a mid-stream failure never touches the caller's table. The fix that removes the asymmetry rather than patching it is a per-input `SAVEPOINT` when the executor is a caller-owned transaction: open it before the input's first statement, `RELEASE` on success, `ROLLBACK TO` and `RELEASE` on failure. That restores the caller's table, removes the partial table, and makes the staging-table special case in `abandonLoad` unnecessary. The same savepoint scope fixes the multi-table-input atomicity gap reported separately for XLSX/ACH/Fedwire.

## Tests to add

- The reproduction above as a regression test in load_into_test.go: `LoadIntoTx` of a path input that fails late under `MalformedRowStop`, asserting inside the still-open transaction that the caller's same-named table keeps its schema and rows and that no partial table remains.
- The reader-input sibling in the same test function, asserting the same invariant, so the two input kinds are pinned to identical failure behavior instead of the current asymmetry.
- A widened sibling: the same path-input failure but with a chunk size larger than the row count (single chunk) and smaller (many chunks), because the partial table's existence depends on how many chunks were inserted before the failure.
- The `LoadInto` (non-Tx) variant stays green: the same failing file through `LoadInto` already leaves the database as it was, and must keep doing so once savepoints are introduced.

nao1215:
## Summary

A load's atomicity was per table rather than per input, so the inputs that map one file onto several tables broke the contract `LoadInto` documents, and a failed path input inside a caller-owned transaction left a partial table where the caller's table had been. Every input now runs inside one scope that a failure undoes: a transaction of its own on a database this package drives, a savepoint inside a transaction the caller owns.

Closes #489
Closes #490

## Changes

- `stream_processor.go`: `runInputScope` wraps each file and each reader input, and `undoInput` returns a caller-owned transaction to the savepoint the input opened and pops it. The rollback runs under `context.WithoutCancel`, since a canceled context is one of the ways a load fails and is no reason to leave the caller holding half an input; a transaction `database/sql` already ended for that same cancellation reports `sql.ErrTxDone`, which is the undo having happened by a larger one.
- `stream_processor.go`: with the scope in place, `loadTable` takes the transaction it runs in and only reports, so `beginLoad` and `abandonLoad` are gone along with the branches that had become unreachable — `beginLoad`'s `*sql.DB` case, `loadTable`'s own commit, and `abandonLoad`'s rollback and staging-table drop.
- `load_into_test.go`: regressions for both issues — a two-sheet workbook whose second sheet is refused, with and without a table it replaces and inside a caller transaction; a path input and a reader input failing late inside `LoadIntoTx`, at one chunk and at many; and the two cancellation halves, one where the caller's transaction outlives the canceled load and one where it was built on the canceled context.
- `stream_processor_error_test.go`: the loading functions now take the transaction a load runs in, so the tests open one through a new `openTestTx` helper, and the two contracts that moved to the scope are exercised through it.
- `example_api_test.go`, `README.md`, `CHANGELOG.md`, `builder.go`: an `ExampleDBBuilder_LoadIntoTx` with its row in the README example index, one CHANGELOG entry per fix, and the guarantee stated in the `LoadIntoTx` godoc.

## Design Decisions

The scope belongs per input rather than per table because an input and a table are not the same thing: a workbook is one input and one table per sheet, an ACH file is one input and several tables, and a Fedwire file is one input with a message table and a write-back row. Putting the scope at the input boundary is what makes the documented sentence true for all of them at once, and it subsumes the staging-table cleanup that `abandonLoad` did by hand for one of the two loading paths.

A savepoint is the only way to give a caller-owned transaction the same guarantee, since that transaction is not this package's to end. Rolling back to the savepoint and releasing it leaves the transaction exactly as the input found it and still usable, which is what a caller who catches the error and commits the rest of their work needs.

The rollback deliberately ignores cancellation. Skipping it when `ctx.Err() != nil` was wrong for the common case where the caller opened their transaction with a different context: the transaction was still alive, and the input's partial table plus the drop of the caller's table survived in it. A regression test pins that case, and it fails against the skipping version.

## Limitations

When the caller's transaction is built on the context the load runs under, cancelling it ends the whole transaction — `database/sql` rolls it back and there is no savepoint left to return to. Nothing of the load reaches the database, which is the right outcome, but the caller's transaction is gone rather than usable. That is `database/sql`'s behavior for a canceled transaction context, not a choice this package makes.

Benchmarks are unchanged: `BenchmarkOpenContext`, `BenchmarkOpenWithAutoSave` and `BenchmarkOpenParquet` move within run-to-run noise over `-count=3`, with `B/op` and `allocs/op` identical.

## Summary by CodeRabbit

* **Bug Fixes**
  * Inputs now load atomically, preventing partial tables or write-back data when loading fails.
  * Failed loads preserve existing tables and caller-managed transaction state.
  * Cancelled loads safely roll back while keeping transactions usable.

* **Documentation**
  * Added guidance and an example for loading data within a caller-managed transaction.


nao1215:
Both review points are addressed in dfc0925.

CodeRabbit's nitpick on the LoadIntoTx godoc: taken, in one sentence rather than two. The godoc now says to begin tx with a context of its own to keep it usable after a failed input, since canceling the context it was begun with ends the whole transaction. Stating the remedy rather than the database/sql mechanism keeps it to what a caller has to do; the mechanism is in the CHANGELOG and pinned by TestLoadIntoTx_CanceledContextTheTransactionWasBuiltOn.

reviewdog's three misspell hits: fixed, along with the two test names that carried the same spelling, so canceled reads the same way throughout the new code.

PR #499, opened 2026-08-24
fix(dialect): translate div, the two-argument trunc and width_bucket

### LINKED ISSUE #498: div, two-argument trunc and width_bucket are refused although both engines define them and SQLite can express them
Found at 95e5f7f while probing the arithmetic family (S6 dialect, filesql-dig-bug session 2026-08-24).

## Phenomenon

Five arithmetic functions that both engines define, and that SQLite can express, are not translated: `div(x, y)`, the two-argument `trunc(x, n)` and `width_bucket(x, lo, hi, n)` under PostgreSQL, and `DIV(x, y)` and the two-argument `TRUNC(x, n)` under GoogleSQL. Each fails with a SQLite message rather than being computed. The failure is loud, so nothing is silently wrong, but the message names something the caller did not write — "no such function: div" for a function PostgreSQL has had since 8.3 — and the same operation is already translated for MySQL, where `TRUNCATE(12.345, 2)` answers 12.34. A caller porting a query across the three dialects finds truncation working in one of them and refused in the other two.

## Reproduction

```go
package filesql_test

import (
	"context"
	"os"
	"path/filepath"
	"testing"

	"github.com/nao1215/filesql"
)

func TestArithmeticFunctionsAreTranslated(t *testing.T) {
	path := filepath.Join(t.TempDir(), "t.csv")
	if err := os.WriteFile(path, []byte("v\n1\n"), 0600); err != nil {
		t.Fatal(err)
	}
	for _, tc := range []struct {
		dialect dialect.Dialect
		query   string
		want    string
	}{
		{dialect.PostgreSQL, `SELECT div(7, 2)`, "3"},
		{dialect.PostgreSQL, `SELECT trunc(12.345, 2)`, "12.34"},
		{dialect.PostgreSQL, `SELECT width_bucket(5.35, 0.024, 10.06, 5)`, "3"},
		{dialect.GoogleSQL, `SELECT DIV(7, 2)`, "3"},
		{dialect.GoogleSQL, `SELECT TRUNC(12.345, 2)`, "12.34"},
	} {
		ctx := context.Background()
		validated, err := filesql.NewBuilder().AddPath(path).WithDialect(tc.dialect).Build(ctx)
		if err != nil {
			t.Fatal(err)
		}
		db, err := validated.Open(ctx)
		if err != nil {
			t.Fatal(err)
		}
		var got string
		if err := db.QueryRow(tc.query).Scan(&got); err != nil {
			t.Errorf("%v: %s: %v", tc.dialect, tc.query, err)
		} else if got != tc.want {
			t.Errorf("%v: %s = %s, want %s", tc.dialect, tc.query, got, tc.want)
		}
		db.Close()
	}
}
```

What filesql answers at 95e5f7f:

```
PostgreSQL  div(7, 2)                            -> ERROR: SQL logic error: no such function: div (1)
PostgreSQL  trunc(12.345, 2)                     -> ERROR: SQL logic error: wrong number of arguments to function trunc() (1)
PostgreSQL  width_bucket(5.35, 0.024, 10.06, 5)  -> ERROR: SQL logic error: no such function: width_bucket (1)
GoogleSQL   DIV(7, 2)                            -> ERROR: SQL logic error: no such function: DIV (1)
GoogleSQL   TRUNC(12.345, 2)                     -> ERROR: SQL logic error: wrong number of arguments to function TRUNC() (1)
MySQL       TRUNCATE(12.345, 2)                  =  12.34
```

The answers the engines give, read from postgres:17-alpine and ghcr.io/goccy/bigquery-emulator:

```
$ docker exec pgprobe2 psql -U postgres -t -c "SELECT div(7,2), div(-7,2), div(7,-2), trunc(12.345,2), trunc(-12.345,2), trunc(12.345,0), width_bucket(5.35,0.024,10.06,5), width_bucket(0.0,0.024,10.06,5), width_bucket(20.0,0.024,10.06,5);"
   3 |  -3 |  -3 | 12.34 | -12.34 |    12 |            3 |            0 |            6
$ docker exec pgprobe2 psql -U postgres -t -c "SELECT div(7,0);"
ERROR:  division by zero

$ curl ... -d '{"query":"SELECT DIV(7,2) AS a, DIV(-7,2) AS b","useLegacySql":false}'
{"rows":[{"f":[{"v":"3"},{"v":"-3"}]}],...}
$ curl ... -d '{"query":"SELECT DIV(7,0) AS c","useLegacySql":false}'
{"error":{...,"message":"failed to scan rows: DIV: zero divided"}}
```

So `div` and `DIV` truncate toward zero, `trunc` truncates toward zero at the given scale, `width_bucket` numbers the buckets from 1 with 0 below the range and n+1 above it, and a zero divisor raises for both spellings of the integer division.

## Expected behavior

The five should compute what the engines compute. None of them needs anything SQLite lacks: an integer division truncating toward zero is the division helper this package just gained, a scale-taking truncation is a multiply, a truncate and a divide, and `width_bucket` is arithmetic on the bounds. The dialect package's doc.go reserves the clear refusal for "known constructs with no SQLite equivalent", and these are not that; they are gaps in the function table, which the same doc says are filled by helpers registered into the driver.

A zero divisor in `div` and `DIV` should raise the way the operators now do, since both engines raise for it. `width_bucket` should refuse a zero-width range, which PostgreSQL does with "lower bound cannot equal upper bound" — worth confirming against the engine before implementing.

## Cause (guess)

This is a guess from reading dialect/functions.go and the call passes. There is nothing wrong in the code; the names are simply absent from the helper table and from the per-dialect call rewrites, so they reach SQLite unchanged and SQLite has no `div` and a one-argument `trunc`. The work is a helper apiece plus a case in `pgRewriteCall` and `googlesqlRewriteCall`, next to the `MOD` cases.

## Tests to add

- The reproduction above in the dialect tests, as a table across the five with the values measured from the engines.
- The signs, in the same table: `div(-7, 2)` and `div(7, -2)` are both -3, and `trunc(-12.345, 2)` is -12.34, so a helper that floors instead of truncating fails.
- The zero divisor: `div(7, 0)` and `DIV(7, 0)` raise, joining the operators.
- The scale boundaries for `trunc`: a scale of 0, a scale larger than the number of decimals, and a negative scale, whose answers should be read from PostgreSQL before they are pinned.
- The three `width_bucket` regions: below the range answers 0, inside answers 1..n, above answers n+1.
- The one-argument forms that already work must keep working: `trunc(12.345)` is 12 and `TRUNC(12.345)` is 12.

nao1215:
## Summary

Five arithmetic functions both engines define reached SQLite unchanged and failed there, although none of them needs anything SQLite lacks and MySQL's equivalent was already translated. `div`, the two-argument `trunc` and `width_bucket` now compute what PostgreSQL computes, and `DIV` and the two-argument `TRUNC` what GoogleSQL computes.

Closes #498

## Changes

- `dialect/operators.go`: `integerDivide`, `truncateScale` and `widthBucket`, which share the operand reader the zero-divisor helpers use.
- `dialect/functions.go`: the three registered under `div`, `trunc_scale` and `width_bucket`.
- `dialect/rewrite.go`, `dialect/postgresql.go`, `dialect/googlesql.go`: `rewriteTruncScaleCall`, which renames only the two-argument `TRUNC` so the one-argument form stays SQLite's, listed as P-22 and G-23.
- `dialect/operators_test.go`: `TestArithmeticFunctionsAreTranslated` over the eighteen values, the signs, the scale boundaries, the three `width_bucket` regions and the refusals.
- `dialect/doc.go`, `CHANGELOG.md`: the record.

## Design Decisions

Every expected value was read from an engine rather than derived. postgres:17-alpine gives `div(7,2)`=3, `div(-7,2)`=-3, `div(7,-2)`=-3, `trunc(12.345,2)`=12.34, `trunc(-12.345,2)`=-12.34, `trunc(12.345,-1)`=10, `trunc(12345.6,-2)`=12300, `width_bucket(5.35,0.024,10.06,5)`=3, `width_bucket(0.0,...)`=0, `width_bucket(20.0,...)`=6 and `width_bucket(5.0,10.0,0.0,5)`=3, and raises for `div(7,0)` and for a range of no width. The BigQuery emulator gives `DIV(7,2)`=3 and `DIV(-7,2)`=-3 and raises for `DIV(7,0)`.

The first implementation of `width_bucket` answered 2 where PostgreSQL answers 3, because the buckets are numbered from 1 and the share of the range names the bucket below. The engine caught it; the arithmetic alone would not have.

`div` and `width_bucket` are registered under their own names rather than dialect-prefixed ones, since both dialects want the same answer and SQLite has neither name. Two-argument `TRUNC` needs the rewrite instead, because SQLite already has a one-argument `trunc` and the two forms have to keep sharing the name, which is how `ROUND` is handled.

## Limitations

GoogleSQL's two-argument `TRUNC` follows PostgreSQL's semantics, which is what BigQuery documents, but is the one value here not confirmed against an engine: the emulator implements only the one-argument form and answers "invalid number of arguments: got 2, want 1".

`width_bucket` refuses a count of zero or less, which PostgreSQL also refuses, but the message is this package's rather than a transcription of PostgreSQL's.

## Summary by CodeRabbit

- **New Features**
  - Added computed support for PostgreSQL and GoogleSQL `div`/`DIV`.
  - Added two-argument `trunc`/`TRUNC`, including negative scales and truncation toward zero.
  - Added PostgreSQL `width_bucket` support, including ascending and descending ranges.
- **Bug Fixes**
  - Zero divisors now consistently return a divide-by-zero error.
  - Invalid bucket counts and ranges are handled correctly.
  - One-argument `trunc` remains SQLite-native.


nao1215:
Taken in de3ff9f, and the engine agrees: postgres:17-alpine answers 0 for `trunc(12.345, -400)`, so a scale whose power of ten underflows truncates the value away rather than keeping it. The other end is unchanged and now has a case of its own — a scale past every decimal the value has overflows the power of ten, and PostgreSQL keeps the value there, which is what `trunc(12.345, 400)` pins.

Issue #414, opened 2026-08-21
A 433-byte Parquet file allocates gigabytes and never returns


nao1215:
Found on 966b647 while fuzzing the binary readers, alongside #413. This one is not a panic: the reader keeps allocating and never returns.

## Phenomenon

A 433-byte Parquet file holds `filesql.OpenContext` forever and grows the heap by roughly 350 MiB per second while it does. Twenty seconds in it has allocated 21 GiB with 7 GiB live, and it is still going; left alone it takes the machine's memory. A program that loads a file someone else supplied — which is what this package is for — is stopped dead by a file smaller than this issue.

Nothing in the file's own metadata looks unusual, which is what makes it hard to defend against from here: it declares 24 rows, three columns, one row group, and column chunks of about three kilobytes. The blow-up happens below that, while a page is being decoded, where this package has no visibility.

A deadline does not help. `pqarrow.FileReader.ReadTable` with a context that expires in two seconds is still running fifteen seconds later, so a caller who bounds their own work cannot bound this, and neither can this package by passing its caller's context down.

## Reproduction

```go
package repro

import (
	"context"
	"encoding/base64"
	"os"
	"path/filepath"
	"runtime"
	"testing"
	"time"

	"github.com/nao1215/filesql"
)

// 433 bytes, found by fuzzing parser.Parse over Parquet.
const slowParquet = "MDAwMBUYFQAVCBUILBUGFRAVMBUwHBhvFTAZQTkEGAYwMDAwMDAVBjAVBCMwGAIwMDcwMDAw" +
	"MDAwMDAVDCMwGAQwMDAwIzBMHDAwMBUKJTAYBTAwMDAwMBYwGRwZMCbaMBwVMBkwMDAwGRgC" +
	"MDAVABYwFtIwFtIBJjAmCBwYIDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwWAgw" +
	"MDAwMDAwMDAZIBUwFTAVMDAVMBUwFTAwMDAmjDAcFTAZMDAwMBkYBDAwMDAVABYwFrIwFrIB" +
	"JrQwJtoBHDYwFjAYBTAwMDAwGAgwMDAwMDAwMDAZIBUwFTAVMDAVMBUwFTAwMDAm3jAcFTAZ" +
[...]

Issue #511, opened 2026-08-25
A 28-byte xz file allocates the 4 GiB dictionary its header asks for


nao1215:
Found on main at 04fd762 by fuzzing the compression codecs (S3), on Linux/amd64 with Go 1.26.

## Phenomenon

An xz file declares the size of the dictionary its decoder must hold, in one byte of the LZMA2 filter's properties, and `github.com/ulikunitz/xz` allocates exactly what the byte asks for before it reads any compressed data. Nothing bounds it, so a 28-byte file whose property byte says 40 — the format's maximum, a 4 GiB dictionary — makes `filesql.Open` reserve 4 GiB and then fail with `unexpected EOF`. Repeating that in one process is where it stops being virtual: twenty `Open` calls on the same 28-byte file take peak resident memory from 17 MiB to 8224 MiB and 2.44 seconds, because the runtime reuses and zeroes the spans. This is the class the Parquet bound closed in #414 — a damaged file must fail in time and memory bounded by its own size — reached through a different codec, and it matters for a long-running process such as sqly's shell, where several files are opened over a session and the cost accumulates.

## Reproduction

```go
package main

import (
	"bytes"
	"context"
	"encoding/binary"
	"fmt"
	"hash/crc32"
	"os"
	"path/filepath"
	"time"

	"github.com/nao1215/filesql"
)

// A minimal xz stream whose LZMA2 filter property asks for the dictionary the
// prop byte encodes: dict = (2 | (prop & 1)) << (prop/2 + 11), so 40 is 4 GiB.
func xzStream(prop byte) []byte {
	var b bytes.Buffer
	b.Write([]byte{0xFD, '7', 'z', 'X', 'Z', 0x00})
	flags := []byte{0x00, 0x00}
	b.Write(flags)
	_ = binary.Write(&b, binary.LittleEndian, crc32.ChecksumIEEE(flags))
	fields := []byte{0x00, 0x00, 0x21, 0x01, prop}
	total := len(fields) + 4 // the CRC is part of Block Header Size
	for total%4 != 0 {
		fields = append(fields, 0x00)
		total = len(fields) + 4
	}
	fields[0] = byte(total/4 - 1)
	b.Write(fields)
	_ = binary.Write(&b, binary.LittleEndian, crc32.ChecksumIEEE(fields))
	b.Write([]byte{0x01, 0x00, 0x00, 'A'}) // a truncated LZMA2 payload
	return b.Bytes()
}

func peakRSSMiB() uint64 {
[...]

PR #513, opened 2026-08-25
fix: bound the working memory an xz or zstd header may declare

### LINKED ISSUE #512: A 14-byte zstd file is served the 512 MiB window its header asks for
Found on main at 04fd762 by fuzzing the compression codecs (S3), on Linux/amd64 with Go 1.26. Same class as the xz dictionary bound, different codec and a different fix.

## Phenomenon

A zstd frame declares the window its decoder must hold, in one byte of the frame header, and `github.com/klauspost/compress/zstd` allocates it before reading any compressed data. filesql passes no decoder options, so it takes the library's permissive default ceiling of 512 MiB: a 14-byte `.zst` file whose window descriptor asks for 512 MiB costs a 513 MiB allocation, and unlike the xz case the file is not damaged — it loads, and `SELECT` returns its one row. Twenty `Open` calls on that file in one process take peak resident memory from 17 MiB to 1050 MiB and 397 ms, against 30 MiB and 17 ms for the same file declaring a 1 MiB window. Nothing in the file says it needs the window; the byte is unchecked and there is no header checksum over it, so what a caller pays is set by a number the file asserts rather than by anything it contains.

## Reproduction

```go
package main

import (
	"bytes"
	"context"
	"fmt"
	"os"
	"path/filepath"
	"time"

	"github.com/nao1215/filesql"
)

// A minimal zstd frame holding "id\n1\n" in one raw block, whose window
// descriptor asks for the window the exponent selects: windowLog = 10 + exp,
// so 19 is 512 MiB and 10 is 1 MiB.
func zstdFrame(exp byte) []byte {
	body := []byte("id\n1\n")
	b := []byte{0x28, 0xB5, 0x2F, 0xFD, 0x00, exp << 3}
	v := uint32(1) | (uint32(len(body)) << 3) // last block, raw
	b = append(b, byte(v), byte(v>>8), byte(v>>16))
	return append(b, body...)
}

func peakRSSMiB() uint64 {
	data, _ := os.ReadFile("/proc/self/status")
	for _, line := range bytes.Split(data, []byte("\n")) {
		if bytes.HasPrefix(line, []byte("VmHWM:")) {
			var kb uint64
			fmt.Sscanf(string(line), "VmHWM: %d kB", &kb)
			return kb >> 10
		}
	}
	return 0
}

func main() {
	dir, _ := os.MkdirTemp("", "zstdwindow")
	defer os.RemoveAll(dir)
	p := filepath.Join(dir, "small.csv.zst")
	if err := os.WriteFile(p, zstdFrame(19), 0o600); err != nil {
		panic(err)
	}
	fi, _ := os.Stat(p)
	fmt.Printf("file: %d bytes, peak RSS before: %d MiB\n", fi.Size(), peakRSSMiB())

	start := time.Now()
	for i := 0; i < 20; i++ {
		db, err := filesql.OpenContext(context.Background(), p)
		if err != nil {
			panic(err) // it loads; that is the point
		}
		_ = db.Close()
	}
	fmt.Printf("after 20 opens: peak RSS %d MiB in %v\n", peakRSSMiB(), time.Since(start))
}
```

```
file: 14 bytes, peak RSS before: 17 MiB
after 20 opens: peak RSS 1050 MiB in 397.429274ms
```

The same program with `zstdFrame(10)`, a 1 MiB window, reports 30 MiB in 16.8 ms. Twenty loads of a 14-byte file that yields one row cost either 1050 MiB or 30 MiB depending only on a byte the file asserts.

## Expected behavior

filesql chooses the ceiling rather than inheriting the library's, and a file asking for more is refused with an error naming the limit instead of being served half a gigabyte. This is the same contract the xz dictionary bound is about and the one #414 was closed on. The ceiling has to clear what real files declare, and measured against the `zstd` CLI on a 27 MB input, `-3` declares a 2 MiB window and `-19` declares 8 MiB, while `--ultra -22` and `--long=27` fall back to a single-segment frame whose window is the content size; the settings that reach a 128 MiB window need both `--ultra -22` or `--long=27` and an input larger than 128 MiB. So 128 MiB reads everything the CLI produces short of `--long=28`, against the 512 MiB in force now.

## Cause (guess)

`internal/codec/codec.go` calls `zstd.NewReader(reader)` with no options, so `decoderOptions.maxWindowSize` keeps its default of `zstd.MaxWindowSize`, which is 1<<29. Unlike the xz case the library does offer a ceiling — `zstd.WithDecoderMaxWindow(n)` — so the fix is passing one rather than pre-validating the header, and a frame beyond it is refused by the library itself with `window size exceeded`, which was confirmed by measurement: the same 14-byte frame under `WithDecoderMaxWindow(8 << 20)` allocated nothing and returned that error. `WithDecoderLowmem(true)` and `WithDecoderConcurrency(1)` were measured too and change nothing here — both still allocated 513 MiB — so neither is a substitute. The error the library returns should be wrapped so a caller sees which file and which limit, rather than a bare `window size exceeded`.

## Tests to add

- In `compression_test.go`: a frame declaring a window past the chosen limit is refused with a classified error naming the limit, and the refusal allocates nothing worth measuring — assert the elapsed time is small, since a regression that restores the allocation shows up there.
- The sibling that pins the boundary from the other side: a frame declaring a window just inside the limit still loads and returns its row, so the limit cannot be tightened into refusing ordinary files.
- In the same file: a round trip through this package's own zstd writer still works, and a file written by the `zstd` CLI at `-19` still loads. Generate both in the test rather than adding fixtures.
- The cross-codec invariant proposed alongside the xz issue covers this one too: for each of the eight codecs, a stream whose header declares an implausible size is refused within a fixed time budget.

nao1215:
Two defects of one class, found by fuzzing the compression codecs. An xz stream and a zstd frame each state in the header how much working memory their decoder must allocate, and both decoders allocate exactly what the header asks for before reading a byte of compressed data, so the cost is set by what the file asserts rather than by anything it contains.

## What changes

A 28-byte xz stream whose LZMA2 property byte says 40 asks for a 4 GiB dictionary; it reserved 4 GiB and then failed with `unexpected EOF`. A 14-byte zstd frame whose window descriptor says 512 MiB allocated 513 MiB and — unlike the xz case, where nothing was damaged — loaded, with `SELECT` returning its one row. Once is mostly virtual; repeating it is not, because the runtime reuses and zeroes the spans. Twenty `Open` calls on each, measured in isolated processes:

| | before | after |
|---|---|---|
| xz declaring 4 GiB | 8224 MiB, 2.44 s | 19 MiB, 2.9 ms |
| zstd declaring 512 MiB | 1050 MiB, 397 ms | 19 MiB, 1.9 ms, refused |
| zstd declaring 1 MiB (control) | 30 MiB, 17 ms | unchanged |

zstd is bounded by `zstd.WithDecoderMaxWindow`, the library's own option, which sees every frame in a file. xz has no such option to pass: `ReaderConfig.DictCap` is a floor rather than a limit, raised to whatever the file declares, and the same 28-byte stream through `DictCap: 1 MiB` still allocated 4096 MiB. So the xz limit is filesql's own, read from the stream header and the first block header before the decoder is built.

## What it does not reach

The xz check covers the first block of the first stream, which is where a damaged file and a plain hostile one declare their size. It does not cover a later block, whose header is not reachable without decoding the blocks before it, nor a second stream concatenated after the first, which the library opens on its own: a 116-byte file made by concatenating an ordinary xz with a 28-byte one declaring 2 GiB still allocates 2056 MiB. So this is a fixed ceiling on the common shape, not the Parquet reader's rule that a file costs no more than its own size, and #511 stays open for the rest. Closing it needs a decoder that takes a dictionary ceiling, which is also the condition for removing the check. The README and the CHANGELOG both say so rather than claiming more than the code delivers.

## Where the limits sit

256 MiB of xz dictionary is four times what `xz -9` declares, against a format maximum of nearly 4 GiB, so that one has room. 128 MiB of zstd window has none: it is exactly what `--ultra -22` and `--long=27` write, and passes only because the comparison is strict, so what it refuses is a deliberately raised `--long=28` or above. Files from `xz -0`, `-6`, `-9`, from `zstd -3`, `-19`, `--ultra -22`, and from this package's own writers all still load, and the tests pin both sides of each boundary.

## Verification

`make test`, `make lint` and `make lint-staticcheck` pass, as does `go test -race ./...`. Cross-compiles for windows/amd64 and darwin/arm64. sqly builds and its whole test suite passes against this branch through a local replace, removed afterwards. Coverage of the new code: `xzWithinDictionaryLimit` and `zstdWithinWindowLimit` 100%, `xzVarint` 100%, `lzma2DictionarySize` 90.6%.

Reading xz also got faster, by more than the check costs: the decoder now reads through the buffer the check leaves in front of it rather than straight from the file. `BenchmarkOpenCompressed/.xz` went from 1836 ms to 729 ms median over five runs with allocations unchanged at 212.5 MB; `.zst` moved from 297.6 ms to 300.1 ms and `.gz`, whose path is untouched, from 394.6 ms to 400.9 ms, which sets the noise floor at about 2%. The benchmark is new in this branch, since the compressed load had none.

Refs #511
Closes #512

## Summary by CodeRabbit

* **Bug Fixes**
  * Added memory safeguards for XZ and Zstandard decompression.
  * Oversized dictionary and window declarations are now rejected with clear errors.
  * Safeguards apply across subsequent blocks and concatenated streams.
  * Damaged or malformed compressed files now fail quickly without excessive memory allocation.

* **Documentation**
  * Documented decompression memory limits and their coverage in the README and changelog.


nao1215:
Thanks — two of the three are fixed in 2ea25d6, and I am leaving the third.

The xz magic check was a real gap and a plain inconsistency with the zstd path beside it, which does compare its magic. Without it, a file named `.xz` that is not an xz stream could have a thirteenth byte reading as a block header size and bytes after it parsing as an LZMA2 filter, and would then be refused for a dictionary it never declared. The regression case is twelve bytes of something else followed by a block header shaped to declare 4 GiB; with the check removed it is refused for 4096 MiB, which is what makes the test worth having rather than one that passes for the wrong reason.

The wall-clock budget is gone too, replaced with bytes allocated, which is what the bound is actually about and what the test above it already uses. The threshold is 300 MiB: above the 256 MiB an accepted header may still legitimately ask for, and far below the 513 MiB and 4096 MiB the unbounded readers took. Both that test and its subtests dropped `t.Parallel()` so `ReadMemStats` counts only their own allocations.

On exporting the sentinel: not doing it. This repository is deliberately shrinking its public surface rather than growing it, and a new exported error has to earn its place by someone needing to branch on it. Nothing does today — sqly, `examples/`, the godoc examples and the README all treat a refused stream the same way they treat any other unreadable one, and the message already names the file's number and the limit for a human reading the failure. If a caller turns up who needs to distinguish a memory-limit refusal from a corrupt stream programmatically, that is the moment to add it, and it can be added compatibly then.

nao1215:
This one is correct and is being left, deliberately. `ReaderConfig.DictCap` really is a floor rather than a ceiling in v0.5.16 — `if dc > config.DictCap { config.DictCap = dc }` in `lzmafilter.go` — and `streamReader.Read` builds a fresh block reader, and therefore a fresh dictionary, per block, with `Reader.Read` opening each concatenated stream in turn. I measured both: a 116-byte file made by concatenating an ordinary `xz -6` stream with a 28-byte one declaring 2 GiB still allocates 2056 MiB after this change.

Inspecting every block is not reachable from where this check sits. A later block's header only appears after the blocks before it have been decoded, so covering it means either interleaving with the decode or reimplementing the block framing, and neither belongs in a peek that runs before the decoder is built. Concatenated streams could be caught by scanning the whole byte stream for the magic, but that pays a full extra pass over every compressed file to close half of the gap while the multi-block half stays open, which is not a trade worth making.

So what this delivers is a fixed ceiling on the shape a damaged file and a plain hostile one actually take — one stream, one block, size declared up front — and not the rule that a file costs no more than its own size. That is stated where someone would rely on it: in `xzWithinDictionaryLimit`'s own comment with the 2056 MiB measurement, in the README, in the CHANGELOG, and in #511, which stays open for the rest and is referenced rather than closed by this PR. Closing it properly needs a decoder that takes a dictionary ceiling, which is the same shape as the Arrow-to-parquet-go swap that closed #414 and is its own piece of work.

PR #516, opened 2026-08-25
ci(examples): build and run the examples against this checkout

### LINKED ISSUE #514: The examples under examples/ are pinned to filesql v0.12.0 and no CI job builds them
Found on main at 4b731d1 while checking what the repository builds.

## Phenomenon

Every directory under `examples/` is its own Go module and all eight of them require `github.com/nao1215/filesql v0.12.0`, thirty-three minor versions behind the current v0.45.0, and nothing under `.github/workflows/` or in the `Makefile` mentions `examples` at all, so the root `go test ./...` never compiles them and no CI job ever has. The README's "Integration examples" table sends readers to these eight directories as the worked demonstration of using filesql with sqlc, GORM, sqlx, Bun, Squirrel and Ent, so the code a newcomer copies is compiled against an API from before most of what CHANGELOG.md records, and the first person to find out that a public symbol moved is that newcomer rather than the pull request that moved it. This also disarms the rule this repository works by, which is that a public symbol is only shrunk after checking `examples/` and sqly for uses of it: the check runs, but it runs against a thirty-three-version-old snapshot of what the examples import, so it cannot report anything.

## Reproduction

```
grep -l 'filesql v0' examples/*/go.mod | xargs grep -h 'nao1215/filesql v'
grep -rn examples .github/workflows Makefile
```

```
	github.com/nao1215/filesql v0.12.0
	github.com/nao1215/filesql v0.12.0
	github.com/nao1215/filesql v0.12.0
	github.com/nao1215/filesql v0.12.0
	github.com/nao1215/filesql v0.12.0
	github.com/nao1215/filesql v0.12.0
	github.com/nao1215/filesql v0.12.0
	github.com/nao1215/filesql v0.12.0
```

The second command prints nothing, which is the defect: there is no job, target or script that builds these modules.

## Expected behavior

The examples should be compiled and run against the checkout that the pull request is changing, so that renaming or unexporting a public symbol fails the pull request instead of rotting a directory the README advertises. Their checked-in `go.mod` should keep naming a released version, because that file is part of what a reader copies, so the wiring to the working tree belongs outside it.

## Cause (guess)

Nothing built them, so nothing kept them current. The `Makefile` drives the root module only (`GO_PKGROOT = ./...`) and no workflow steps outside it.

## Tests to add

- A script, `examples/verify.sh`, that for each example creates a temporary `go.work` naming the repository root and that example, then runs `go vet ./...` and `go run .` under it, so the example compiles against the working tree while its `go.mod` stays untouched.
- A workflow with one job per example so a failure names the integration that broke, on Linux and one Go version, since these exercise the public API rather than platform behavior and the unit test matrix already covers three operating systems.
- A `make examples` target so the same check runs locally before a public symbol is shrunk.

### LINKED ISSUE #515: Two example READMEs document output the example does not print
Found on main at 4b731d1 by running each example and diffing its output against the "Expected Output" block of its own README.

## Phenomenon

Two of the eight example READMEs document output the example has never printed. `examples/multi-format/README.md` shows the JOIN total for the Keyboard row as `239.97`, but `price` is inferred REAL and `3 * 79.99` in float64 is `239.96999999999997`, which is what the program prints. `examples/squirrel/README.md` shows the `age >= 30` result ordered `Bob 35` then `Diana 31` although the query is `OrderBy("age")`, and omits Alice from the `LIKE '%a%'` result although SQLite's LIKE is case-insensitive for ASCII and matches her. These blocks are the only statement of what the example does for a reader who is not going to run it, so a wrong block teaches a wrong expectation about the library: the multi-format one hides that a CSV number column becomes REAL and carries float arithmetic, and the squirrel one suggests LIKE is case-sensitive here. `examples/multi-format/README.md` also has a "Supported Formats" list naming Parquet and XLSX, and the root README's table row calls it "Join across CSV, TSV, LTSV, and Parquet", while `main.go` opens only `products.csv`, `orders.tsv` and `logs.ltsv`.

## Reproduction

```
cd examples/squirrel && go run .
```

```
=== Users Age >= 30 ===
id	name	age
4	Diana	31
2	Bob	35

=== Users with 'a' in name ===
id	name
1	Alice
3	Charlie
4	Diana
```

The README for the same two sections lists Bob before Diana and does not list Alice.

## Expected behavior

The block should hold what the program prints. Where the two disagree the program is right in both cases here: 31 sorts before 35, and `Alice` contains an `a` under a case-insensitive LIKE. The description of an example should also name what the example does rather than what the library supports, so the multi-format entry should say CSV, TSV and LTSV.

## Cause (guess)

The blocks were written by hand rather than pasted from a run, and nothing ever compared them again.

## Tests to add

- The check belongs with the one that builds the examples: after running each example, diff its output against the fenced block that follows `## Expected Output` in its README, so the README is the golden file and there is no separate testdata to drift away from it.
- Report a missing or empty `## Expected Output` block as a failure too, so a new example cannot join the directory undocumented.

nao1215:
## What

The eight modules under `examples/` all required filesql v0.12.0, thirty-three minor versions behind the current release, and nothing in `.github/workflows/` or the `Makefile` named `examples`, so no job has ever compiled them. This bumps them to v0.45.0 and adds `examples/verify.sh` plus a workflow with one job per example, so from here a change to the public API is checked against the code the README's "Integration examples" table sends people to.

## How

`examples/verify.sh` creates a temporary `go.work` naming the repository root and one example, then runs `go vet ./...` and `go run .` under it. Wiring it through the workspace rather than `go mod edit -replace` leaves the checked-in `go.mod` showing the released version a reader would actually import, and leaves no diff behind after a local run. The script then diffs what the example prints against the fenced block after `## Expected Output` in that example's README, which makes the README the golden file instead of a fixture that can drift away from it, and reports a missing block as a failure so a new example cannot join the directory undocumented. `make examples` runs all eight, and CONTRIBUTING now points at it for changes to the public API.

Running it found two README blocks that had never been true. `multi-format` documented the Keyboard row's JOIN total as 239.97 where the REAL column's arithmetic gives 239.96999999999997, and `squirrel` documented the `age >= 30` rows in the wrong order against its own `OrderBy("age")` and left Alice out of a `LIKE '%a%'` that matches her, since SQLite's LIKE is case-insensitive for ASCII. Both are corrected to what the programs print. The multi-format example's format list and the README's row for it also claimed Parquet and XLSX while `main.go` opens only CSV, TSV and LTSV; both now name what the example does.

## Verification

All eight examples vet, run and match their README under `make examples`. To confirm the job catches what it exists for, `OpenContext` was unexported locally and the basic example failed to vet with `undefined: filesql.OpenContext`, which is the failure a future API shrink should produce. `shellcheck` and `actionlint` are clean on the new files, and the root test suite and README sync tests still pass.

Closes #514
Closes #515

## Summary by CodeRabbit

* **New Features**
  * Added automated validation for building, vetting, running, and comparing all example applications with their documented output.
  * Added CI checks for supported examples on pull requests, pushes, and manual runs.

* **Documentation**
  * Updated contribution guidance with example validation instructions.
  * Clarified formats used by the multi-format example and refreshed expected outputs.
  * Updated integration examples and changelog entries to reflect verified behavior.


nao1215:
All three applied in 0119ff6: the workflow now declares `permissions: contents: read`, verify.sh rejects an Expected Output block that holds only whitespace rather than only an empty one, and the changelog sentence is rewritten.

Issue #527, opened 2026-08-25
A .z file holding two zlib streams loads only the first, with no error


nao1215:
Found on main at 3f60bb4 while re-checking every codec for what a concatenated stream costs.

## Phenomenon

A `.z` file holding more than one zlib stream loads only the first one, with no error. Ten concatenated streams of the same 16-byte payload are 280 bytes and produce 16 bytes of output and a nil error, so nine tenths of the file is dropped and nothing says so. The same test over the other codecs reads every stream: ten concatenated gzip members give 160 bytes, and so do xz, zstd, snappy, s2 and lz4. A caller who runs `cat a.csv.z b.csv.z > both.csv.z` — which works for `.gz`, where multi-member files are part of the format — gets a table holding the first file's rows and no sign that the second file's rows are missing.

Silence is the problem rather than the truncation. Two zlib streams in one file are not something the zlib format defines, so refusing the second is a defensible answer; answering with a table that is missing most of its rows is not, because a caller cannot tell that from a small file.

## Reproduction

```go
func TestConcatenatedZlib(t *testing.T) {
	var one bytes.Buffer
	w, closeFn, err := codec.ZLIB.NewWriter(&one)
	require.NoError(t, err)
	_, err = io.WriteString(w, "id,name\n1,alice\n")
	require.NoError(t, err)
	require.NoError(t, closeFn())

	var many []byte
	for range 10 {
		many = append(many, one.Bytes()...)
	}

	reader, closeReader, err := codec.ZLIB.NewReader(bytes.NewReader(many))
	require.NoError(t, err)
	out, readErr := io.ReadAll(reader)
	require.NoError(t, closeReader())

	t.Logf("%d bytes in, %d bytes out, err=%v", len(many), len(out), readErr)
}
```

```
280 bytes in, 16 bytes out, err=<nil>
```

The same shape over gzip prints `400 bytes in, 160 bytes out, err=<nil>`.

## Expected behavior

Either every stream is read, the way gzip's members are, or the bytes behind the first stream are reported. Reading them is the friendlier answer and matches what a caller who concatenated the files expects, and it is what the xz reader in this package now does. Reporting them is the smaller change. Either way the result stops being a table that quietly lost most of its rows.

## Cause (guess)

[...]

PR #528, opened 2026-08-25
fix(xz): check every stream's dictionary, not only the first one

nao1215:
## What

The xz dictionary bound added in #511 covered the first stream of a file and no more. An xz file may hold several streams one after another — what `cat a.xz b.xz` produces, and what a rotated log archive looks like — and the decoder opened the later ones itself, past anything this package had looked at. An ordinary xz with a 28-byte stream declaring a 4 GiB dictionary appended to it is 100 bytes, and it decoded to seventeen bytes and allocated 4104 MiB on the way, returning a nil error where the refusal should have been. The measured figures are 520, 2056 and 4104 MiB for declarations of 512 MiB, 2 GiB and 4 GiB; all three now cost 8 MiB and are refused by name.

## How

`xzStreams` opens the streams one at a time and runs the dictionary check in front of each. The split rests on the decoder reading a single stream on request: asked for one, it reads past the end to see whether anything follows, and fails when something does. Those bytes are the next stream's, so a `pushbackReader` puts back whatever the failing read took — whatever it took, rather than a fixed count, so this does not depend on how the library chooses to look — and the next stream is opened only when what comes back really is a stream header, with the format's four-zero-byte padding stepped over. A decoder that read past the end for its own reasons therefore still surfaces its own error.

## What does not change

Two genuine streams still read as one file, with or without padding between them. One stream reads as itself. A truncated stream is still an error rather than a short read. Bytes behind the last stream that are not a stream are still an error, now reported as unexpected data after a stream rather than as an invalid header magic.

What remains outside the check is a later block within a stream, whose header is not reachable without decoding the blocks before it. Closing that still needs a decoder that takes a dictionary ceiling, which stays the condition for removing this check; `xz.ReaderConfig.DictCap` is not it, since lzmafilter.go raises it to whatever the file declares.

## Verification

`TestXZChecksEveryStream` covers the hostile stream behind an ordinary one with and without padding, two genuine streams with and without padding, one stream alone, trailing bytes that are not a stream, and a truncated stream, asserting both the bytes read and that a refusal costs under 64 MiB. `TestPushbackReader` covers the piece it rests on. `BenchmarkOpenCompressed/.xz` measured 752/733/729 ms here against 716/707/725 ms on main, with `B/op` and `allocs/op` unchanged — about 2% for a check that now runs per stream. The whole suite and `golangci-lint` are clean, and sqly builds and passes its tests against this branch.

The same measurement over gzip, zstd, zlib, snappy, s2 and lz4 found their concatenated cost flat and bounded by their formats, with one exception filed as #527: a `.z` file holding several zlib streams reads only the first, with no error.

Refs #511

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection when reading concatenated XZ streams by enforcing dictionary-size limits on every stream.
  * Reduced the risk of excessive memory allocation from malformed or oversized compressed data.

* **Documentation**
  * Updated the changelog to clarify memory limits for files containing multiple XZ streams.


nao1215:
Applied in ffac495. A file that ends with the stream padding the format allows now ends there rather than reporting unexpected data: the padding is stepped over and the source running out is taken as the end of the file, but only when padding is what ran out, so a stream cut short still surfaces the decoder's error instead of reading as a whole file. Two cases cover it, one padding group and three.

PR #530, opened 2026-08-25
fix(load): wait out another load's lock instead of failing on it

### LINKED ISSUE #529: Concurrent LoadInto fails instead of waiting, and README names only one of the two errors
Found on main at 3f60bb4 by running LoadInto from several goroutines against each shape a caller's database can have.

## Phenomenon

Two loads into the same database at the same time leave one of them failing, and README says so and tells the caller to hold their own lock. What README does not say is how badly it fails or how many ways: sixteen concurrent `LoadInto` calls against a file database left the database empty and sixteen errors behind, and the error is not the one README names. A file database answers `filesql: database operation failed: failed to commit transaction: database is locked (5) (SQLITE_BUSY)`, while `database schema is locked: main (262)` is what a shared-cache database answers — including the one `Open` returns. Measured with four, eight and sixteen goroutines, each loading its own CSV into one database:

| goroutines | `:memory:` pinned to one connection | shared cache | file | the database `Open` returns |
|---|---|---|---|---|
| 4 | 0 of 4 failed | 0 of 4 | 3 of 4 | 2 of 4 |
| 8 | 0 of 8 | 1 of 8 | 6 of 8 | 3 of 8 |
| 16 | 0 of 16 | 10 of 16 | 16 of 16 | 11 of 16 |

A caller who reads README's Concurrency section learns that the database is safe to share and that loading is the exception, then meets an error whose text appears nowhere in filesql and names something they did not write. Holding a lock around every load is a contract this library can keep on its own: SQLite refuses a second writer rather than queueing it, and waiting the lock out is what every application built on SQLite does about that.

## Reproduction

```go
func TestConcurrentLoadInto(t *testing.T) {
	db, err := sql.Open("sqlite", filepath.Join(t.TempDir(), "load.db"))
	require.NoError(t, err)
	defer db.Close()

	dir := t.TempDir()
	const goroutines = 8
	paths := make([]string, goroutines)
	for i := range goroutines {
		paths[i] = filepath.Join(dir, fmt.Sprintf("t%d.csv", i))
		require.NoError(t, os.WriteFile(paths[i], []byte("id,name\n1,alice\n"), 0o600))
	}

	var wg sync.WaitGroup
	errs := make([]error, goroutines)
	for i := range goroutines {
		wg.Go(func() { errs[i] = filesql.LoadInto(t.Context(), db, paths[i]) })
	}
	wg.Wait()
	for i, err := range errs {
		require.NoErrorf(t, err, "load %d", i)
	}
}
```

```
Error: Received unexpected error:
  /tmp/.../t0.csv: filesql: database operation failed: failed to commit transaction: database is locked (5) (SQLITE_BUSY)
Messages: load 0
```

## Expected behavior

A load that meets another load's lock should wait for it and start over, within a budget, rather than failing at once. That is what the lock is for, it is what the drivers' `busy_timeout` exists to express, and it removes a rule the caller currently has to remember. Where waiting cannot help — an input that cannot be read a second time, since starting over would have nothing left to read — the limit should be stated in README rather than left to be discovered.

## Cause (guess)

`runInputScope` in stream_processor.go begins a transaction, runs the load and commits, and returns whatever error any of the three gives. Nothing distinguishes an error that says "someone else holds this" from one that says "this input is wrong", so the first is reported as if it were the second.

## Tests to add

- `TestConcurrentLoadInto` in concurrency_test.go, over the three shapes above and over the database this package opens, asserting every load succeeded and every table has its row. Run under `-race`.
- The reader case as its own test, since a reader is spent by the attempt that reads it and only the steps before the reading can be tried again.
- The negative: an error that is not a lock is still returned at once rather than retried for five seconds, so a bad input does not become a slow bad input.

nao1215:
## What

Creating a table takes a write lock, and SQLite refuses a second holder rather than queueing it, so a load that met another load's lock came back at once with its table not created. Sixteen concurrent `LoadInto` calls against a file database left the database empty and sixteen errors behind; eleven of sixteen failed against the database `Open` returns. The message was the driver's rather than this package's, and there were two of them — `database is locked (5) (SQLITE_BUSY)` from a file database, `database schema is locked: main (262)` from a shared-cache one — while README named only the second and told the caller to hold their own lock around every load.

| goroutines | `:memory:` pinned | shared cache | file | the database `Open` returns |
|---|---|---|---|---|
| 4 | 0 of 4 failed | 0 of 4 | 3 of 4 | 2 of 4 |
| 8 | 0 of 8 | 1 of 8 | 6 of 8 | 3 of 8 |
| 16 | 0 of 16 | 10 of 16 | 16 of 16 | 11 of 16 |

All of them now land.

## How

`runInputScope` splits into `runInputTx`, which does one attempt, and a retry around it. Beginning the transaction is retried whatever the input is, since nothing has been read yet; the whole attempt is retried when the input can be read again, which is a path or a directory. The wait doubles from a millisecond to fifty and carries jitter, because the loads that collide are the loads that would otherwise retry in step, and the budget is five seconds — where the drivers' `busy_timeout` defaults sit — after which the database's own error comes back unchanged. The context ends the waiting too.

What counts as a lock is read from the driver's result code rather than from the message, because the message differs between a file database and a shared-cache one and is not this package's to depend on: `SQLITE_BUSY` and `SQLITE_LOCKED`, matched on the low byte so the extended codes above them — 262 for a shared-cache table, 517 for a stale snapshot — are covered by the same test.

A reader passed to `AddReader` is the one input that cannot be started over, since its bytes are gone once read. Only the steps before the reading are retried for it, which is enough for a database this package opened and not always enough for a file database; README says so rather than leaving it to be found.

`LoadIntoTx` is untouched: the transaction is the caller's, and one transaction belongs to one goroutine.

## Verification

`TestConcurrentLoadInto` runs eight goroutines against a pinned `:memory:` pool, a file database and the database `Open` returns, checking every load succeeded and every table has its row; `TestConcurrentLoadIntoFromReaders` does the same for readers. `TestLoadRetryOnlyWaitsForLocks` pins the other half: the four lock codes are waited on, the five that mean the input is wrong are not, and a duplicate-column CSV still fails in well under a second rather than being waited on for five. The suite is clean under `-race`, `golangci-lint` is clean, `BenchmarkOpenContext` is unchanged at 319/320/320 ms against 327/314/316 ms on main with identical allocations, and sqly builds and passes its tests against this branch.

Closes #529

## Summary by CodeRabbit

* **Bug Fixes**
  * Concurrent loads into the same database now wait and retry when SQLite is temporarily locked.
  * Lock retries use bounded backoff and stop promptly for non-lock errors or canceled operations.
  * Reader-based inputs are handled safely without attempting to reread consumed data.

* **Documentation**
  * Updated concurrency guidance, including retry behavior and limitations.
  * Clarified that database builders should be created separately for concurrent operations.


nao1215:
Applied in 73dad0e. The wait is now cut to what is left of the budget, and a budget already spent returns the lock error without another attempt, so the last try happens inside the five seconds rather than just past them.

PR #537, opened 2026-08-25
perf(xlsx): find a sheet's date cells in its own XML

### LINKED ISSUE #531: Loading an 18 MB workbook costs 1.8 GB, and one call inside the date normalization is 1.3 GB of it
Found on main at ec97202 by measuring each format's load in a process of its own and then breaking the XLSX read into stages.

## Phenomenon

An 18.5 MB workbook of 200,000 rows costs 1,843 MB of peak resident memory to load, against 100 MB for the same rows as a 32.8 MB CSV and 177 MB as a 32.2 MB Parquet file. README's "Memory and streaming" section tells a caller to budget about twice the file's size, which is right for CSV and wrong by a factor of fifty for XLSX. A workbook of a few hundred megabytes cannot be loaded on an ordinary machine, and nothing says so before the attempt.

Most of it is not the reading. Breaking the same load into stages, each in a process of its own:

| stage | peak RSS |
|---|---|
| open the workbook | 23.9 MB |
| GetRows | 267.3 MB |
| GetRows with RawCellValue | 276.2 MB |
| one GetCellStyle call | 1618.7 MB |
| one GetCellValue call | 1604.2 MB |
| NormalizeXLSXDates over the rows | 1646.3 MB |
| the whole load | 1843.4 MB |

A single random-access call into the workbook costs 1.34 GB, because it makes excelize unmarshal the whole worksheet into its object model, while the streaming row read costs 267 MB. `NormalizeXLSXDates` makes one such call per non-empty cell — 2.4 million of them for this sheet — so the first of them buys the whole cost and the rest are paid for already. Date normalization, not row reading, is what makes a workbook expensive.

## Reproduction

Write 200,000 rows to a workbook, then in a fresh process for each stage, open it and stop after the named call, reporting `VmHWM` from /proc/self/status.

```go
f, err := excelize.OpenFile(path)          // 23.9 MB
rows, err := f.GetRows("data")             // 267.3 MB
_, err = f.GetCellStyle("data", "A2")      // 1618.7 MB
```

The third line is one call about one cell.

## Expected behavior

A workbook should cost what the rows in it cost, in the way a CSV does. The row read already does: 267 MB for an 18.5 MB file is in the same country as CSV's 100 MB for 32.8 MB, allowing for XLSX being compressed. What is out of proportion is a per-cell lookup that materializes the sheet.

## Cause (guess)

`NormalizeXLSXDates` in internal/reader/xlsx_date.go decides whether a cell holds a date from its style, and reads the style with `f.GetCellStyle(sheet, axis)` and the serial with `f.GetCellValue(sheet, axis, excelize.Options{RawCellValue: true})`. Both are random access, and excelize serves random access from an unmarshalled worksheet, which it builds on the first such call and keeps. `GetRows` does not go through it: it decodes the sheet XML as a stream.

The options, none of them obviously free:

- Decide the date columns without random access, by reading each cell's style id in the same streaming pass the rows come from. excelize's `Rows` iterator hands back rendered values and no style, so this means reading `xl/worksheets/sheetN.xml` here, which is duplicating what the library does.
- Ask `xl/styles.xml` whether the workbook defines any date-shaped number format at all, and skip normalization when it does not. That is cheap and covers an analytics export with no dates in it, and does nothing for a workbook that has dates.
- Compare `GetRows()` against `GetRows(RawCellValue: true)`: a cell whose rendering differs from its stored value is formatted, which is a necessary condition for a date. It is not a sufficient one — a thousands separator on a plain number also differs — so it narrows the set of cells that need a style lookup rather than removing the lookups.
- Leave it and state the cost in README, which is what this issue does in the meantime.

No issue is filed with excelize, per this repository's rule.

## Tests to add

- Whatever is chosen, a benchmark-tagged measurement in memory_footprint_test.go: `TestLoadMemoryFootprintByFormat` already loads each format in a process of its own and reports peak resident memory, so the figure to hold is the XLSX row of that table.
- The behavior that must not move while the cost does: a date cell becomes ISO 8601, a number formatted as a number does not, text that looks like a date does not, an elapsed duration does not, and a 1904-epoch workbook converts against 1904. Those cases are in internal/reader/xlsx_date_test.go and in parser/xlsx_test.go.

## Removal condition

If excelize grows a way to read a cell's style during a streaming row read, the workaround comes out and that is used instead.

nao1215:
## What

The half of #531 the previous change left open. A workbook that holds no date format was already cheap; one with a date-formatted cell still cost 1939 MB for 18.5 MB of file, because deciding which cells wear that style meant asking the library about them one at a time, and the first such question makes excelize unmarshal the whole worksheet into its object model.

| 18.5 MB workbook, 200,000 rows | before this round | after the style-table check | now |
|---|---|---|---|
| no date format | 1849 MB | 351 MB | 422 MB |
| one date-formatted cell | 1939 MB | 1939 MB | 390 MB |

The two now cost the same, which is what they should: a date cell is a number and a style, not extra rows.

## How

The date cells are found by walking the sheet's own XML. The loader already has the workbook's bytes — the format is a zip and they were buffered to open it at all — so the sheet part is located through the workbook's relationships (`xl/workbook.xml` for the sheet's relationship id, `xl/_rels/workbook.xml.rels` for what it points at) rather than guessed from a position: a workbook whose sheets were reordered or deleted does not name its parts in order, and the test fixtures are exactly that shape, since they delete `Sheet1`.

The walk records only the cells that are dates, so what it holds is the date cells rather than the sheet. A cell counts when its type is a number — a shared string or an inline string is text whatever its style says — and its style is one of the date styles, which the style table gives cheaply.

The decision itself did not move. `isoFromSerial` is now `isoFromRaw` with the same body and a thin wrapper for the exported path, so both ways convert through the same code: the built-in and custom date formats, the colour brackets that are not elapsed units, the 1904 epoch, and the serials below 61 that name no day.

`parser.NormalizeXLSXDates` keeps asking cell by cell. It is handed an open workbook and no bytes, so that is all it can do, and it stays as the fallback for a workbook whose parts do not say where the sheet is — a sheet missing its dates would be worse than a slow one.

## Verification

`TestReadXLSXDates` is new and covers the normalization where a load meets it, which nothing did before: six formats that are dates and four that are not, text that only looks like one, the phantom February 29, 1900, a 1904-epoch workbook, and a date column beside a string column and a number column. `TestScanSheetDates` covers the walk on shapes a writer other than excelize produces — a row or a cell with no reference, which follows the one before — and `TestCellRef` covers the reference itself up to XFD1048576. `TestDateCellsFromXML` covers finding the sheet, a sheet that is not there, and bytes that are not an archive.

The suite is clean under `-race`, `golangci-lint` is clean, `BenchmarkOverwriteWorkbook` is unchanged at 214/208/220 ms and 180 MB/op, and sqly builds and passes its tests against this branch.

Closes #531

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced peak memory usage when loading XLSX workbooks, including files containing date-formatted cells.
  * Updated benchmark results to reflect approximately 422 MB peak memory for a 17.6 MB workbook.

* **Bug Fixes**
  * Improved recognition and conversion of spreadsheet dates, including custom formats and the 1904 date system.
  * Preserved non-date values and text that resembles dates during workbook reading.

* **Tests**
  * Added coverage for date detection, conversion, invalid workbook data, and mixed date/non-date sheets.


nao1215:
Applied in 5dcc6e1. The `Workbook.data` comment said 351 MB, which was the figure before this change and only for a workbook without dates; it now says 422 MB against 1939 MB, which is what this leaves behind. The earlier changelog entry that filed this issue also claimed it was left unfixed, and now points at the entries that fixed it.

PR #556, opened 2026-08-25
fix: wait for a load's lock the way the godoc says, and say when a context ended the wait

### LINKED ISSUE #554: The lock budget is charged for parsing, so a workbook load gives up after three attempts
Found on main (e38e9de) while looking into a Windows CI failure of TestConcurrentLoadInto, then measured on Linux.

## Phenomenon

The five seconds that bound how long a load waits for another connection to let go are wall clock over the whole attempt, and an attempt is a whole input. For a CSV that costs nothing, because the `DROP TABLE` / `CREATE TABLE` come before the rows are read, so the budget really is spent waiting. For an input that has to be parsed before the load knows what tables to make, it is the opposite: a workbook is opened and its sheets listed on every attempt, so three attempts at a 40000-row XLSX consume the whole budget and the load reports `database is locked (5) (SQLITE_BUSY)` having waited a fraction of a second. `(*DBBuilder).LoadInto`'s godoc says "A load that meets another load's lock waits it out", so a caller has no reason to expect that how long their load waits depends on how expensive their file is to parse.

## Reproduction

Hold the database from another connection for eight seconds, then load a workbook into it. The holder lets go at eight seconds, so a load that waits gets in.

```go
package main

import (
	"context"
	"database/sql"
	"fmt"
	"os"
	"path/filepath"
	"time"

	"github.com/nao1215/filesql"
	"github.com/xuri/excelize/v2"
	_ "modernc.org/sqlite"
)

func main() {
	dir, _ := os.MkdirTemp("", "xlsxprobe")
	defer os.RemoveAll(dir)

	f := excelize.NewFile()
	sheet := f.GetSheetName(0)
	for r := range 40000 {
		cell, _ := excelize.CoordinatesToCellName(1, r+1)
		f.SetCellValue(sheet, cell, r)
		cell, _ = excelize.CoordinatesToCellName(2, r+1)
		f.SetCellValue(sheet, cell, fmt.Sprintf("name-%d", r))
	}
	path := filepath.Join(dir, "book.xlsx")
	f.SaveAs(path)
	f.Close()

	dbPath := filepath.Join(dir, "load.db")
	holder, _ := sql.Open("sqlite", dbPath)
	defer holder.Close()
	ctx := context.Background()
	holder.ExecContext(ctx, "CREATE TABLE keep(x)")
	tx, _ := holder.BeginTx(ctx, nil)
	tx.ExecContext(ctx, "INSERT INTO keep VALUES (1)")
	time.AfterFunc(8*time.Second, func() { tx.Rollback() })

	db, _ := sql.Open("sqlite", dbPath)
	defer db.Close()

	start := time.Now()
	err := filesql.LoadInto(ctx, db, path)
	fmt.Printf("elapsed=%v err=%v\n", time.Since(start).Round(time.Millisecond), err)
}
```

```
elapsed=5.001s err=/tmp/xlsxprobe3025515037/book.xlsx: sheet Sheet1: filesql: database operation failed: failed to create table: database is locked (5) (SQLITE_BUSY)
```

It gives up at 5.001s, which is the budget to the millisecond, after three attempts that were parsing rather than waiting. The same probe with a one-row CSV instead of the workbook makes about 140 attempts in those five seconds, which is what the budget is supposed to buy.

## Expected behavior

The five seconds bound the waiting, so a load waits the same amount whatever its input costs to parse, and the workbook above gets in when the lock is free at eight seconds. Waiting is not the same as retrying, though: each attempt at an expensive input pays for its parse again, and a hundred attempts at a workbook that takes a second to read is a load that looks hung, so the whole thing wants a wall-clock bound of its own -- around half a minute, long enough for any queue worth waiting for and short enough to be an answer rather than a hang.

This does not make an arbitrarily long queue of loads succeed, and it should not be expected to: thirty-two goroutines loading a 200000-row CSV each into one file database is about ten seconds of serialized writing, so the ones at the back exhaust five seconds of honest waiting and fail. That is SQLite refusing a second writer, not this bug. What is in scope here is that the budget is spent on the right thing.

## Cause (guess)

`retryWhileLocked` in `stream_processor.go` sets `deadline := time.Now().Add(loadLockBudget)` once and then measures `time.Until(deadline)` around a `step()` that is the whole load of one input (`runInputTx` -> `streamFileToDatabase`). Nothing separates the sleeping from the attempt, so the cost of the attempt is charged to the waiting. Counting the sleeps down instead of the clock is the small version of the fix; the reason an attempt costs so much in the first place is that the transaction begins deferred and the first statement that needs the write lock cannot be issued until the input has been read far enough to know the schema, which is a larger change and not what this asks for.

## Tests to add

- A test in `concurrency_test.go` that gives the retry a step which sleeps longer than the whole budget and answers with a real `SQLITE_BUSY` for its first few attempts, then succeeds: it has to reach the success, which it cannot if the sleeping in the step is charged to the budget. The budget has to be a parameter of an unexported helper for this, so the test does not wait five seconds; do not add a public knob.
- A test that a lock nobody lets go of still comes back, bounded, with the database's own error.
- A test that a failure which is not a lock is returned on the first attempt, so nothing that is not contention becomes a wait.

### LINKED ISSUE #555: A load canceled while waiting out a lock reports SQLITE_BUSY instead of the context error
Found on main (e38e9de) while probing the concurrency surface of LoadInto.

## Phenomenon

A load whose context expires while it is waiting out another connection's lock reports the lock error and not the context error, so `errors.Is(err, context.DeadlineExceeded)` and `errors.Is(err, context.Canceled)` are both false for an operation that ended for exactly that reason. A caller that gives a load a deadline and branches on it, which is the usual shape for a server handler, cannot tell "the deadline you set ran out" from "the input is bad" and has no way to reach the context error at all, since the returned error chain does not carry it. Every load path is affected, `OpenContext` included, because the retry sits under all of them.

## Reproduction

```go
package main

import (
	"context"
	"database/sql"
	"errors"
	"fmt"
	"os"
	"path/filepath"
	"time"

	"github.com/nao1215/filesql"
	_ "modernc.org/sqlite"
)

func main() {
	dir, _ := os.MkdirTemp("", "cancelprobe")
	defer os.RemoveAll(dir)

	path := filepath.Join(dir, "people.csv")
	os.WriteFile(path, []byte("id,name\n1,alice\n"), 0o600)
	dbPath := filepath.Join(dir, "load.db")

	// Hold the database from another connection for the whole run.
	holder, _ := sql.Open("sqlite", dbPath)
	defer holder.Close()
	holder.ExecContext(context.Background(), "CREATE TABLE keep(x)")
	tx, _ := holder.BeginTx(context.Background(), nil)
	tx.ExecContext(context.Background(), "INSERT INTO keep VALUES (1)")
	defer tx.Rollback()

	db, _ := sql.Open("sqlite", dbPath)
	defer db.Close()

	ctx, cancel := context.WithTimeout(context.Background(), 300*time.Millisecond)
	defer cancel()
	err := filesql.LoadInto(ctx, db, path)
	fmt.Printf("error: %v\n", err)
	fmt.Printf("errors.Is(err, context.DeadlineExceeded) = %v\n", errors.Is(err, context.DeadlineExceeded))
	fmt.Printf("ctx.Err() = %v\n", ctx.Err())
}
```

```
error: /tmp/cancelprobe1789079088/people.csv: filesql: database operation failed: failed to create table: database is locked (5) (SQLITE_BUSY)
errors.Is(err, context.DeadlineExceeded) = false
ctx.Err() = context deadline exceeded
```

The load ended because its context ended, and the error it returns says nothing about that.

## Expected behavior

A load that stops because its context ended returns an error whose chain carries `ctx.Err()`, so `errors.Is(err, context.DeadlineExceeded)` and `errors.Is(err, context.Canceled)` answer for it, which is what the standard library and the rest of this package do with a context. The lock error is still worth keeping, since it says what the load was waiting for, so the two belong joined rather than one replacing the other.

## Cause (guess)

`retryWhileLocked` in `stream_processor.go` returns the last step error unchanged when the context ends during the backoff sleep:

```go
select {
case <-ctx.Done():
	timer.Stop()
	return err
case <-timer.C:
}
```

Nothing there consults `ctx.Err()`, so the reason the wait stopped is dropped. The same holds for the budget-exhausted return a few lines above, but that one is not a context failure and should stay as it is.

## Tests to add

- A test in `concurrency_test.go` that runs the retry helper with a context canceled during the wait and a step that always answers `SQLITE_BUSY`, asserting both `errors.Is(err, context.Canceled)` and that the lock error is still reachable in the chain.
- The deadline sibling of the same test, so the two context errors are both pinned rather than one standing in for the other.
- A test that a context which is already done before the first attempt still runs that attempt and reports its result, so an already-expired context does not turn into a different error than the operation itself would give.

nao1215:
Two defects in how a load waits for another connection's lock, found by digging the concurrency surface after a Windows CI run of TestConcurrentLoadInto failed with SQLITE_BUSY. Closes #554, Closes #555.

## The budget was charged for parsing

The five seconds that bound how long one load waits for another to let go were wall clock over the whole attempt, and an attempt is a whole input. A CSV pays nothing for that, since the drop and the create come before its rows are read -- about 140 attempts fit in the five seconds, which is what the budget is meant to buy. A workbook is the opposite: it is opened and its sheets listed before the load knows what tables to make, so three attempts at a 40000-row XLSX used the budget up and the load failed with `database is locked` having waited a fraction of a second.

Measured against a lock held by another connection for eight seconds, one 40000-row XLSX load: `main` gives up at 5.001s with SQLITE_BUSY, this branch gets in at 8.05s when the lock is free.

What is counted down is now the sleeping. Retrying is not waiting, though, and each attempt at an expensive input pays for its parse again, so the whole retry carries a wall-clock bound of half a minute as well -- long enough for any queue of loads worth waiting for, short enough to be an answer rather than a hang.

This does not make an arbitrarily long queue of loads succeed, and it is not meant to: 32 goroutines loading a 200000-row CSV each into one file database is about ten seconds of serialized writing, so the ones at the back still exhaust five seconds of honest waiting and fail. That is SQLite refusing a second writer. The godoc of `(*DBBuilder).LoadInto` said a load "waits it out", which promises more than SQLite can give, so it now says the wait is bounded.

## Cancellation was reported as a lock

While waiting, the retry returned the lock error and nothing else once the context was done, so `errors.Is(err, context.DeadlineExceeded)` and `errors.Is(err, context.Canceled)` were both false for an operation that ended for exactly that reason, and a caller who gave the load a deadline could not tell their own deadline from a bad file. Every load path is affected, `OpenContext` included, since the retry sits under all of them. The context error is now joined to the lock error rather than replacing it, because the lock is what the load was waiting for and the only thing that says why the deadline was not enough.

## Verification

- Both regression groups fail on `main` and pass here; `retryWhileLocked` and `retryWhileLockedFor` are at 100% statement coverage.
- The tests drive a real SQLITE_BUSY from a second connection rather than a hand-built error, since the driver does not export a way to make one.
- `make test`, `make lint`, `make lint-staticcheck` green; `-race` green over the concurrency tests; `GOOS=windows` and `GOOS=darwin` cross-build.
- sqly builds and its tests pass against this branch through a local replace.
- No benchmark: the retry runs on contention, and the one thing added to the uncontended path is a single `time.Now()` per input.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database lock retries by tracking waiting time accurately, preventing expensive inputs from prematurely exhausting the retry budget.
  - Capped lock-wait retries at five seconds of waiting and thirty seconds overall.
  - Preserved cancellation and deadline errors alongside lock errors for clearer diagnostics.
- **Documentation**
  - Clarified `LoadInto` behavior when waiting for database locks.


nao1215:
Both findings applied in 67957a0. The select now leaves the context check to a single place after it, so a canceled load cannot pay for one more attempt when both cases are ready, and the godoc names the wall-clock bound alongside the waiting budget.

PR #575, opened 2026-08-25
fix(dialect): follow the source dialect for the function names SQLite shares

### LINKED ISSUE #568: LOG answers a base-10 logarithm under MySQL and GoogleSQL, and reads its arguments in the wrong order under GoogleSQL
Found on main (v0.46.0) while diffing the three dialects against mysql:8.4, postgres:17-alpine and the BigQuery documentation.

## Phenomenon

LOG reaches SQLite untranslated, so it takes SQLite's meaning rather than the source dialect's: SQLite's one-argument log(X) is the base-10 logarithm and its two-argument log(B, X) takes the base first. MySQL's LOG(X) and GoogleSQL's LOG(X) are the natural logarithm, and GoogleSQL's LOG(X, Y) is the logarithm of X to base Y, so it takes the base second. A caller who writes LOG under either dialect gets a plausible number that is silently wrong: LOG(10) answers 1 instead of 2.302585092994046, and under GoogleSQL LOG(8, 2) answers 0.3333333333333334 instead of 3. PostgreSQL is unaffected because its log() means what SQLite's means in both arities.

## Reproduction

```go
package main

import (
	"context"
	"fmt"
	"strings"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
	for _, c := range []struct {
		d    dialect.Dialect
		expr string
		want string
	}{
		{dialect.MySQL, "LOG(10)", "2.302585092994046"},
		{dialect.GoogleSQL, "LOG(10)", "2.302585092994046"},
		{dialect.GoogleSQL, "LOG(8, 2)", "3"},
		{dialect.PostgreSQL, "log(10)", "1"},
		{dialect.PostgreSQL, "log(2, 8)", "3"},
	} {
			AddReader(strings.NewReader("id\n1\n"), "t", filesql.FileTypeCSV).
			WithDialect(c.d).Build(ctx)
		if err != nil {
			panic(err)
		}
		conn, err := db.Open(ctx)
		if err != nil {
			panic(err)
		}
		var got any
		err = conn.QueryRowContext(ctx, "SELECT "+c.expr+" FROM t").Scan(&got)
		conn.Close()
		fmt.Printf("%-11s %-12s got %-22v want %s (err %v)\n", c.d, c.expr, got, c.want, err)
	}
}
```

```
mysql       LOG(10)      got 1                      want 2.302585092994046 (err <nil>)
googlesql   LOG(10)      got 1                      want 2.302585092994046 (err <nil>)
googlesql   LOG(8, 2)    got 0.33333333333333337    want 3 (err <nil>)
postgresql  log(10)      got 1                      want 1 (err <nil>)
postgresql  log(2, 8)    got 3                      want 3 (err <nil>)
```

The two MySQL and GoogleSQL one-argument calls answer a base-10 logarithm where the dialect means a natural one, and the GoogleSQL two-argument call answers the logarithm of 2 to base 8 where the dialect means the logarithm of 8 to base 2.

## Expected behavior

Under MySQL, LOG(X) is LN(X) and LOG(B, X) is the logarithm of X to base B, which is already what SQLite's two-argument log means, so only the one-argument form needs rewriting. Under GoogleSQL, LOG(X) is a synonym of LN(X) and LOG(X, Y) is the logarithm of X to base Y, so the one-argument form needs rewriting and the two-argument form needs its arguments swapped. Reference values read from the engines: MySQL answers 2.302585092994046 for LOG(10) and 3 for LOG(2, 8); PostgreSQL answers 1 for log(10) and 3 for log(2, 8); the BigQuery documentation for LOG states that with one argument it is a synonym of LN and with two it computes the logarithm of X to base Y. LOG2 and LOG10 already agree with SQLite in every dialect and should stay untranslated. A domain error (LOG(0), LOG(-1)) stays NULL, which is what SQLite answers and what MySQL answers; BigQuery raises there, and that difference is out of scope for this issue.

## Cause (guess)

There is no LOG entry in dialect/mysql.go's mysqlRewriteCall or dialect/googlesql.go's googlesqlRewriteCall, so the call is passed through to SQLite's own log(). The fix belongs next to the other per-dialect renames in those two files: rewrite MySQL LOG(x) to ln(x), rewrite GoogleSQL LOG(x) to ln(x), and rewrite GoogleSQL LOG(x, y) to log(y, x) so the base reaches SQLite first. Both rewrites have to recurse into their arguments the way the neighboring rewriteRenameCall calls do.

## Tests to add

- In dialect/mysql_test.go, a translation case asserting LOG(x) becomes ln(x) and LOG(b, x) is left alone, plus an execution case in dialect/functions_test.go TestUDFExecution asserting SELECT LOG(10) answers 2.302585092994046 after translation.
- In dialect/googlesql_test.go, translation cases for both arities (LOG(x) to ln(x), LOG(8, 2) to log(2, 8)), and an execution case in TestGoogleSQLUDFExecution asserting the value 3.
- A sibling case for the spellings that must not change: LOG10(100) is 2 and LOG2(8) is 3 under every dialect, and PostgreSQL log(10) stays 1. This is the case that would catch a fix that rewrites too much.
- A case asserting the rewrite is stable when LOG appears inside another call (for example ROUND(LOG(10), 3)), since the fix has to recurse into arguments.

### LINKED ISSUE #569: to_hex under the PostgreSQL dialect hexes its argument's text instead of converting the integer
Found on main (v0.46.0) while diffing the PostgreSQL dialect against postgres:17-alpine.

## Phenomenon

The dialect helper functions are registered on the SQLite driver under their bare names for every dialect at once, so PostgreSQL's to_hex(bigint) resolves to the GoogleSQL helper fnToHex, which hexes the bytes of its argument's text. to_hex(255) answers "323535" (the hex of the three characters "255") where PostgreSQL answers "ff". The result is a plausible string of hex digits, so a caller reading a query's output has no way to notice the value is wrong; only its length gives it away.

## Reproduction

```go
package main

import (
	"context"
	"fmt"
	"strings"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
		AddReader(strings.NewReader("id\n1\n"), "t", filesql.FileTypeCSV).
		WithDialect(dialect.PostgreSQL).Build(ctx)
	if err != nil {
		panic(err)
	}
	conn, err := db.Open(ctx)
	if err != nil {
		panic(err)
	}
	defer conn.Close()
	var got any
	if err := conn.QueryRowContext(ctx, "SELECT to_hex(255) FROM t").Scan(&got); err != nil {
		panic(err)
	}
	fmt.Printf("to_hex(255) = %v, want ff\n", got)
}
```

```
to_hex(255) = 323535, want ff
```

The helper hexed the characters "255" instead of converting the integer 255 to its hexadecimal digits.

## Expected behavior

Under the PostgreSQL dialect, to_hex(n) converts an integer to its lowercase hexadecimal representation, with no 0x prefix and no leading zeros: postgres:17-alpine answers "ff" for to_hex(255), "0" for to_hex(0), and "ffffffffffffffff" for to_hex(-1) because the argument is read as a 64-bit two's complement value. PostgreSQL has no string form of the function at all, so a non-integer argument should raise rather than fall back to hexing the text. GoogleSQL's TO_HEX(bytes) keeps its current meaning: BigQuery only accepts BYTES there and reports a type error for TO_HEX(255), so the two dialects need two helpers rather than one shared one. MySQL is unaffected: HEX() is already routed to mysql_hex.

## Cause (guess)

dialect/functions.go registers "to_hex" once, bound to fnToHex, which is written for GoogleSQL's BYTES semantics, and dialect/postgresql.go has no entry for to_hex in its call rewriter, so the PostgreSQL query reaches the GoogleSQL helper. The fix is the pattern the package already uses for LPAD, RPAD, SUBSTR, CAST, DIVIDE and MOD: register a postgresql_to_hex helper implementing the integer conversion and rewrite to_hex to it in postgresqlRewriteCall, leaving to_hex bound to fnToHex for GoogleSQL.

## Tests to add

- In dialect/functions_test.go TestPostgreSQLUDFExecution, cases for to_hex(255) = "ff", to_hex(0) = "0" and to_hex(-1) = "ffffffffffffffff".
- In dialect/functions_test.go TestGoogleSQLUDFExecution, a case pinning the GoogleSQL meaning that must not change: TO_HEX of a byte string still answers the hex of its bytes.
- A sweep test asserting that every helper name registered in dialect/functions.go which is also a real function name in another supported dialect resolves to that dialect's meaning. This is the class of bug: a bare helper name shared by two dialects that mean different things by it. Candidates worth asserting alongside to_hex are left, right, regexp_replace, greatest, least, initcap and md5.

### LINKED ISSUE #570: REGEXP_REPLACE reads MySQL's position argument as PostgreSQL flags, and replaces every match where PostgreSQL replaces one
Found on main (v0.46.0) while diffing the MySQL and PostgreSQL dialects against mysql:8.4 and postgres:17-alpine.

## Phenomenon

One helper implements REGEXP_REPLACE for all three dialects, and it implements PostgreSQL's signature only partly and MySQL's not at all. Two things go wrong. Under MySQL, REGEXP_REPLACE(subject, pattern, replacement, pos) takes a start position as its fourth argument, but the helper reads that argument as a PostgreSQL flag string: REGEXP_REPLACE('aaa', 'a', 'X', 2) answers "Xaa" where MySQL answers "aXX", and the five-argument form MySQL documents (pos and occurrence) is rejected with "REGEXP_REPLACE expects 3 or 4 arguments". Under PostgreSQL, the three-argument form replaces every match where PostgreSQL replaces only the first: regexp_replace('aaa', 'a', 'X') answers "XXX" where PostgreSQL answers "Xaa". Both are silent wrong answers on a function callers reach for when cleaning text, and the MySQL one turns a positional argument into a no-op flag string.

## Reproduction

```go
package main

import (
	"context"
	"fmt"
	"strings"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
	for _, c := range []struct {
		d    dialect.Dialect
		expr string
		want string
	}{
		{dialect.PostgreSQL, "regexp_replace('aaa', 'a', 'X')", "Xaa"},
		{dialect.PostgreSQL, "regexp_replace('aaa', 'a', 'X', 'g')", "XXX"},
		{dialect.MySQL, "REGEXP_REPLACE('aaa', 'a', 'X')", "XXX"},
		{dialect.MySQL, "REGEXP_REPLACE('aaa', 'a', 'X', 2)", "aXX"},
		{dialect.MySQL, "REGEXP_REPLACE('aaa', 'a', 'X', 1, 2)", "aXa"},
	} {
			AddReader(strings.NewReader("id\n1\n"), "t", filesql.FileTypeCSV).
			WithDialect(c.d).Build(ctx)
		if err != nil {
			panic(err)
		}
		conn, err := db.Open(ctx)
		if err != nil {
			panic(err)
		}
		var got any
		err = conn.QueryRowContext(ctx, "SELECT "+c.expr+" FROM t").Scan(&got)
		conn.Close()
		fmt.Printf("%-11s %-40s got %-8v want %-6s (err %v)\n", c.d, c.expr, got, c.want, err)
	}
}
```

```
postgresql  regexp_replace('aaa', 'a', 'X')          got XXX      want Xaa    (err <nil>)
postgresql  regexp_replace('aaa', 'a', 'X', 'g')     got XXX      want XXX    (err <nil>)
mysql       REGEXP_REPLACE('aaa', 'a', 'X')          got XXX      want XXX    (err <nil>)
mysql       REGEXP_REPLACE('aaa', 'a', 'X', 2)       got Xaa      want aXX    (err <nil>)
mysql       REGEXP_REPLACE('aaa', 'a', 'X', 1, 2)    got <nil>    want aXa    (err SQL logic error: dialect: REGEXP_REPLACE expects 3 or 4 arguments, got 5 (1))
```

The MySQL four-argument call ignored the start position and replaced the first match instead of the matches at and after position 2, the MySQL five-argument call was rejected outright, and the PostgreSQL three-argument call replaced every match instead of one.

## Expected behavior

Under MySQL, REGEXP_REPLACE(subject, pattern, replacement[, pos[, occurrence[, match_type]]]) replaces every match from character position pos onward, or only the occurrence-th match from there when occurrence is given and non-zero; pos is 1-based and mysql:8.4 answers "aXX" for pos 2 and "aXa" for pos 1 with occurrence 2. Under PostgreSQL, regexp_replace(source, pattern, replacement[, flags]) replaces the first match only unless the flags contain "g"; postgres:17-alpine answers "Xaa" without flags and "XXX" with "g". Under GoogleSQL, REGEXP_REPLACE(value, regexp, replacement) has three arguments and replaces every match, which is what the helper does today. The current code comment in fnRegexpReplace records the shared-helper behavior as a deliberate simplification, but the package now routes per-dialect meanings through per-dialect helpers (mysql_lpad, postgresql_substr, googlesql_mod), so that reason no longer holds and the comment should go with the change.

## Cause (guess)

dialect/functions.go registers a single "regexp_replace" bound to fnRegexpReplace, which reads a fourth argument as PostgreSQL flags and caps the arity at four, and neither dialect/mysql.go nor dialect/postgresql.go rewrites the call. The fix is a mysql_regexp_replace helper accepting three to six arguments with pos, occurrence and a match_type string, and a postgresql_regexp_replace helper that defaults to replacing the first match only, each wired in its dialect's call rewriter. GoogleSQL keeps the existing helper.

## Tests to add

- In dialect/functions_test.go TestUDFExecution, MySQL cases for the four-argument pos form ("aXX"), the five-argument occurrence form ("aXa"), an occurrence of 0 meaning every match from pos, a pos past the end of the subject answering the subject unchanged, and a pos of 0 or a negative pos being rejected the way MySQL rejects it.
- In dialect/functions_test.go TestPostgreSQLUDFExecution, cases for the three-argument form replacing one match, the "g" flag replacing all, the "i" flag still folding case, and a back-reference (\1) still expanding in both the one-match and the all-matches path.
- In dialect/functions_test.go TestGoogleSQLUDFExecution, a case pinning that the three-argument GoogleSQL form still replaces every match, so the fix does not move GoogleSQL onto PostgreSQL's default.
- A NULL case per dialect in TestUDFNullHandling: a NULL in any argument answers NULL rather than raising.

### LINKED ISSUE #571: FORMAT(x, d) under the MySQL dialect answers the number unchanged instead of a rounded, grouped string
Found on main (v0.46.0) while diffing the MySQL dialect against mysql:8.4.

## Phenomenon

MySQL's FORMAT(X, D) rounds a number to D decimal places and returns it as a string with thousands separators. SQLite has a function of the same name (an alias of printf), so the call is passed through and the first argument is read as a format string with no verbs in it: FORMAT(1234.5678, 2) answers "1234.5678" and the second argument is dropped. The caller gets back the number they passed in, unrounded and ungrouped, with no error to say the request was not honored.

## Reproduction

```go
package main

import (
	"context"
	"fmt"
	"strings"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
		AddReader(strings.NewReader("id\n1\n"), "t", filesql.FileTypeCSV).
		WithDialect(dialect.MySQL).Build(ctx)
	if err != nil {
		panic(err)
	}
	conn, err := db.Open(ctx)
	if err != nil {
		panic(err)
	}
	defer conn.Close()
	for _, expr := range []string{"FORMAT(1234.5678, 2)", "FORMAT(1234.5678, 0)", "FORMAT(-1234.5, 1)"} {
		var got any
		err := conn.QueryRowContext(ctx, "SELECT "+expr+" FROM t").Scan(&got)
		fmt.Printf("%-22s got %-12v (err %v)\n", expr, got, err)
	}
}
```

```
FORMAT(1234.5678, 2)   got 1234.5678    (err <nil>)
FORMAT(1234.5678, 0)   got 1234.5678    (err <nil>)
FORMAT(-1234.5, 1)     got -1234.5      (err <nil>)
```

mysql:8.4 answers "1,234.57", "1,235" and "-1,234.5" for those three; every call here answered the printf expansion of the first argument instead.

## Expected behavior

Under the MySQL dialect, FORMAT(X, D) rounds X to D decimal places, groups the integer part in threes with commas, and returns a string: mysql:8.4 answers "1,234.57" for FORMAT(1234.5678, 2), "1,235" for D = 0, and "-1,234.5" for FORMAT(-1234.5, 1). A negative D is treated as 0. The third, locale argument MySQL accepts is out of scope: it should be refused with the package's unsupported-syntax error rather than silently ignored, since ignoring it would answer en_US formatting for a query that asked for something else. GoogleSQL's FORMAT is a printf-style function and keeps its current mapping to SQLite printf; PostgreSQL's format() is also printf-style and is likewise unaffected. The package already has the grouping logic this needs: dialect/aggregate.go carries a groupThousands helper covered by TestGroupThousands.

## Cause (guess)

dialect/mysql.go's mysqlRewriteCall has no FORMAT entry, so the call reaches SQLite's own format(). The fix is a mysql_format helper registered in dialect/functions.go and a rewrite in mysqlRewriteCall, refusing an arity other than two the way other MySQL-specific rewrites refuse forms they cannot represent.

## Tests to add

- In dialect/mysql_test.go, a translation case asserting FORMAT(x, d) becomes mysql_format(x, d) and that a three-argument call is refused with ErrUnsupportedSyntax.
- In dialect/functions_test.go TestUDFExecution, execution cases for "1,234.57", "1,235", "-1,234.5", a value below 1000 that gets no separator, a value with several groups (1234567.891 with D = 2 is "1,234,567.89"), D larger than the value's precision (padding zeros), and a NULL argument answering NULL.
- A sibling case pinning that GoogleSQL FORMAT and PostgreSQL format() still expand printf verbs, so the fix does not follow the name across dialects.

### LINKED ISSUE #572: LEFT and RIGHT answer PostgreSQL's result for a negative length under MySQL and GoogleSQL
Found on main (v0.46.0) while diffing all three dialects against mysql:8.4, postgres:17-alpine and the BigQuery emulator.

## Phenomenon

LEFT and RIGHT are implemented by one shared helper that follows PostgreSQL's rule for a negative length: it removes that many characters from the far end. MySQL and GoogleSQL do not agree. MySQL answers an empty string for a negative length, and BigQuery raises an error. So LEFT('abcd', -1) answers "abc" under every dialect, where MySQL answers "" and BigQuery reports that the length must not be negative. A negative length usually comes from arithmetic on a column (LEFT(name, n - 1) with n = 0), so a caller gets a nearly-complete string back where their engine gives them an empty one or an error.

## Reproduction

```go
package main

import (
	"context"
	"fmt"
	"strings"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
	for _, c := range []struct {
		d    dialect.Dialect
		expr string
		want string
	}{
		{dialect.MySQL, "LEFT('abcd', -1)", "empty string"},
		{dialect.MySQL, "RIGHT('abcd', -1)", "empty string"},
		{dialect.GoogleSQL, "LEFT('abcd', -1)", "an error"},
		{dialect.PostgreSQL, "left('abcd', -1)", "abc"},
		{dialect.PostgreSQL, "right('abcd', -1)", "bcd"},
	} {
			AddReader(strings.NewReader("id\n1\n"), "t", filesql.FileTypeCSV).
			WithDialect(c.d).Build(ctx)
		if err != nil {
			panic(err)
		}
		conn, err := db.Open(ctx)
		if err != nil {
			panic(err)
		}
		var got any
		err = conn.QueryRowContext(ctx, "SELECT "+c.expr+" FROM t").Scan(&got)
		conn.Close()
		fmt.Printf("%-11s %-20s got %-6q want %-12s (err %v)\n", c.d, c.expr, fmt.Sprint(got), c.want, err)
	}
}
```

```
mysql       LEFT('abcd', -1)     got "abc"  want empty string (err <nil>)
mysql       RIGHT('abcd', -1)    got "bcd"  want empty string (err <nil>)
googlesql   LEFT('abcd', -1)     got "abc"  want an error     (err <nil>)
postgresql  left('abcd', -1)     got "abc"  want abc          (err <nil>)
postgresql  right('abcd', -1)    got "bcd"  want bcd          (err <nil>)
```

The MySQL and GoogleSQL rows took PostgreSQL's meaning for the negative length; only the PostgreSQL rows are right.

## Expected behavior

PostgreSQL's left(str, n) and right(str, n) with a negative n drop |n| characters from the other end, which is what the helper does today and should keep doing. MySQL's LEFT(str, n) and RIGHT(str, n) answer an empty string for a negative n; mysql:8.4 answers "" for LEFT('abcd', -1). GoogleSQL raises for a negative length: the BigQuery emulator reports "LEFT: unexpected length value. length must be positive number", and the BigQuery documentation states the length must not be negative, so the dialect should raise an error rather than answer a string. A length of zero answers an empty string in all three, and a length past the end of the string answers the whole string in all three; neither changes.

## Cause (guess)

dialect/functions.go registers one "left" and one "right" bound to leftRight, whose comment says it implements PostgreSQL's negative-count semantics, and neither dialect/mysql.go nor dialect/googlesql.go rewrites the call. The fix is the per-dialect helper pattern already used for LPAD and SUBSTR: mysql_left and mysql_right clamping a negative length to zero, googlesql_left and googlesql_right raising on a negative length, both wired in their dialect's call rewriter, with the existing helper left for PostgreSQL.

## Tests to add

- In dialect/functions_test.go TestUDFExecution and TestGoogleSQLUDFExecution, cases for a negative length in each dialect: MySQL answers "", GoogleSQL raises, PostgreSQL answers "abc".
- Sibling boundary cases in the same table, per dialect: length 0 answers "", length equal to the string length answers the whole string, length greater than the string length answers the whole string, and a multi-byte string is counted in characters rather than bytes (LEFT('日本語', 2) is "日本").
- A NULL case in TestUDFNullHandling for each new helper.

### LINKED ISSUE #573: A fractional or exponent string cast to an integer rounds instead of raising (PostgreSQL, GoogleSQL) or truncating (MySQL)
Found on main (v0.46.0) while diffing all three dialects against mysql:8.4, postgres:17-alpine and the BigQuery emulator.

## Phenomenon

castToInt falls back to ParseFloat when a string does not parse as an integer, so a fractional or exponent string becomes a rounded integer under every dialect. None of the three engines does that. PostgreSQL raises "invalid input syntax for type integer" for CAST('1.5' AS integer); BigQuery raises for CAST('1.5' AS INT64) and answers NULL for SAFE_CAST; MySQL answers 1, because its string-to-integer conversion stops at the first character that cannot be part of an integer rather than reading the whole number. filesql answers 2 in all three cases, and 1000 for CAST('1e3' AS SIGNED) where MySQL answers 1. A caller loading a CSV whose column holds "1.5" therefore gets a value that is neither their engine's answer nor their engine's error.

## Reproduction

```go
package main

import (
	"context"
	"fmt"
	"strings"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
	for _, c := range []struct {
		d    dialect.Dialect
		expr string
		want string
	}{
		{dialect.MySQL, "CAST('1.5' AS SIGNED)", "1"},
		{dialect.MySQL, "CAST('1e3' AS SIGNED)", "1"},
		{dialect.MySQL, "CAST('12abc' AS SIGNED)", "12"},
		{dialect.PostgreSQL, "CAST('1.5' AS integer)", "an error"},
		{dialect.GoogleSQL, "CAST('1.5' AS INT64)", "an error"},
		{dialect.GoogleSQL, "SAFE_CAST('1.5' AS INT64)", "NULL"},
	} {
			AddReader(strings.NewReader("id\n1\n"), "t", filesql.FileTypeCSV).
			WithDialect(c.d).Build(ctx)
		if err != nil {
			panic(err)
		}
		conn, err := db.Open(ctx)
		if err != nil {
			panic(err)
		}
		var got any
		err = conn.QueryRowContext(ctx, "SELECT "+c.expr+" FROM t").Scan(&got)
		conn.Close()
		fmt.Printf("%-11s %-28s got %-6v want %-9s (err %v)\n", c.d, c.expr, got, c.want, err)
	}
}
```

```
mysql       CAST('1.5' AS SIGNED)        got 2      want 1         (err <nil>)
mysql       CAST('1e3' AS SIGNED)        got 1000   want 1         (err <nil>)
mysql       CAST('12abc' AS SIGNED)      got 12     want 12        (err <nil>)
postgresql  CAST('1.5' AS integer)       got 2      want an error  (err <nil>)
googlesql   CAST('1.5' AS INT64)         got 2      want an error  (err <nil>)
googlesql   SAFE_CAST('1.5' AS INT64)    got 2      want NULL      (err <nil>)
```

Every row rounded the string's numeric value; MySQL should have truncated at the first non-digit and the two strict dialects should have refused the string.

## Expected behavior

PostgreSQL and GoogleSQL read a string cast to an integer type as an integer literal and raise when it is not one: postgres:17-alpine reports invalid input syntax for '1.5', and BigQuery reports the same class of error, with SAFE_CAST answering NULL. So under those two dialects the cast should fail with ErrInvalidCast for '1.5', '1e3' and any other string whose text is not an integer, and googlesql_safe_cast should answer NULL for the same input. MySQL keeps its coercion but has to follow its own rule: mysql:8.4 answers 1 for CAST('1.5' AS SIGNED), 1 for CAST('1e3' AS SIGNED), 12 for CAST('12abc' AS SIGNED), -12 for CAST('-12abc' AS SIGNED), 0 for CAST('.5' AS SIGNED) and clamps '9223372036854775808' to the maximum, so the leading run it reads is an optional sign followed by digits, with no point and no exponent. Note that MySQL's numeric-context coercion is a different rule and reads the whole number ('1e3' + 0 is 1000); only CAST to an integer type truncates, so the existing numericPrefix behavior is still right for the contexts that use it and should not be changed for both.

## Cause (guess)

dialect/cast.go castToInt, after ParseInt fails, calls ParseFloat and hands the result to roundForDialect, which rounds for all three dialects; the strict flag only decides what happens for a value out of range, not for a string that names a non-integer. The fix is to split the string path by dialect: for the strict dialects, fail with ErrInvalidCast when the text is not an integer literal (keeping the existing out-of-range branch, which must still answer the range error rather than a parse error), and for MySQL take the leading sign-and-digits run instead of the float value.

## Tests to add

- In dialect/cast_test.go TestCastSemantics, MySQL cases for '1.5' to 1, '1e3' to 1, '-12abc' to -12, '.5' to 0, and ' 12 ' to 12 so the existing whitespace trim is kept.
- In dialect/cast_test.go TestCastRejectsInvalidValues, PostgreSQL and GoogleSQL cases for '1.5' and '1e3' failing with ErrInvalidCast, and a case asserting googlesql_safe_cast answers NULL for the same strings.
- Sibling cases that must not change: a numeric (not string) 1.5 still rounds to 2 for MySQL and GoogleSQL and to 2 for PostgreSQL under TestRoundForDialect's rules, an integer string still casts, and the out-of-range string '9223372036854775808' still clamps for MySQL and raises the range error for the strict dialects (TestCastStringPastTheFloatRange covers the neighboring case).

### LINKED ISSUE #574: FORMAT('%t', x) and FORMAT('%T', x) answer NULL under the GoogleSQL dialect
Found on main (v0.46.0) while diffing the GoogleSQL dialect against the BigQuery documentation and the BigQuery emulator.

## Phenomenon

GoogleSQL FORMAT is rewritten to SQLite printf, which covers the C verbs the two share but not the two verbs BigQuery adds: %t, which prints a value in its compact string form, and %T, which prints it as the SQL literal that would produce it. SQLite's printf does not know either verb, and a printf call whose format holds an unknown verb answers NULL, so FORMAT('%t', true) and FORMAT('%T', 'x') answer NULL instead of "true" and "\"x\"". A caller building a display string or a log line gets a NULL row rather than text, and the NULL propagates through any concatenation it feeds.

## Reproduction

```go
package main

import (
	"context"
	"fmt"
	"strings"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
		AddReader(strings.NewReader("id\n1\n"), "t", filesql.FileTypeCSV).
		WithDialect(dialect.GoogleSQL).Build(ctx)
	if err != nil {
		panic(err)
	}
	conn, err := db.Open(ctx)
	if err != nil {
		panic(err)
	}
	defer conn.Close()
	for _, expr := range []string{`FORMAT('%t', true)`, `FORMAT('%T', 'x')`, `FORMAT('%t', 1.5)`, `FORMAT('%d items', 3)`} {
		var got any
		err := conn.QueryRowContext(ctx, "SELECT "+expr+" FROM t").Scan(&got)
		fmt.Printf("%-22s got %-10v (err %v)\n", expr, got, err)
	}
}
```

```
FORMAT('%t', true)     got <nil>      (err <nil>)
FORMAT('%T', 'x')      got <nil>      (err <nil>)
FORMAT('%t', 1.5)      got <nil>      (err <nil>)
FORMAT('%d items', 3)  got 3 items    (err <nil>)
```

The three calls using BigQuery's own verbs answered NULL; only the C verb worked.

## Expected behavior

BigQuery documents %t as the value's printable form and %T as the value's literal form: FORMAT('%t', true) is "true", FORMAT('%T', 'x') is "\"x\"" (a quoted string literal), FORMAT('%t', 1.5) is "1.5", FORMAT('%T', 1.5) is "1.5", and a NULL argument prints "NULL" for both verbs. The types filesql carries are the SQLite ones, so the mapping is: an integer prints as its digits for both; a real prints as its shortest round-trip form for both; a string prints bare for %t and double-quoted with the usual escapes for %T; NULL prints as NULL for both; a boolean reaching the helper as the integer 0 or 1 cannot be told from an integer and prints as digits, which is a limit worth stating in the same place the other dialect limits are recorded rather than guessing at the caller's intent. Refusing the two verbs with the package's unsupported-syntax error would also be acceptable and is better than the current NULL, but printing them costs little more than refusing them.

## Cause (guess)

dialect/googlesql.go's googlesqlRewriteCall maps FORMAT to printf with rewriteRenameCall, so the whole format string, verbs and all, is handed to SQLite. The fix is a googlesql_format helper that walks the format string, handles %t and %T itself, and delegates the rest of the run to fmt.Sprintf with the arguments it consumed, so the C verbs keep the behavior they have now.

## Tests to add

- In dialect/functions_test.go TestGoogleSQLUDFExecution, cases for %t and %T over each type filesql can carry: an integer, a real, a string with a quote or a backslash in it, and NULL.
- Cases mixing verbs in one format string ('%s=%T') and a literal '%%' next to a %t, which is the boundary a hand-written scanner gets wrong.
- A sibling case pinning that the C verbs still work after the change (%d, %s, %05.2f), so the fix does not regress what printf already handled.
- A case for a %t with no argument left to consume, asserting it fails or prints the way the rest of the helper decides rather than panicking.

nao1215:
This fixes seven silent wrong answers found by running the same expressions through filesql's three dialects and through the engines they name: mysql:8.4, postgres:17-alpine, and the BigQuery emulator alongside the BigQuery documentation. Every expected value in the tests was read from an engine rather than written from memory.

## What was wrong

All seven come from one shape: a function name SQLite already has, meaning something else there than in the dialect the query was written in. The call ran, so nothing failed; it just answered a different function's value.

| Query | Answered | Engine answers |
|---|---|---|
| `LOG(10)` (MySQL, GoogleSQL) | 1 | 2.302585092994046 |
| `LOG(8, 2)` (GoogleSQL) | 0.3333333333333334 | 3 |
| `to_hex(255)` (PostgreSQL) | `323535` | `ff` |
| `regexp_replace('aaa', 'a', 'X')` (PostgreSQL) | `XXX` | `Xaa` |
| `REGEXP_REPLACE('aaa', 'a', 'X', 2)` (MySQL) | `Xaa` | `aXX` |
| `FORMAT(1234.5678, 2)` (MySQL) | `1234.5678` | `1,234.57` |
| `LEFT('abcd', -1)` (MySQL) | `abc` | empty string |
| `LEFT('abcd', -1)` (GoogleSQL) | `abc` | an error |
| `CAST('1.5' AS SIGNED)` (MySQL) | 2 | 1 |
| `CAST('1.5' AS integer)` (PostgreSQL, GoogleSQL) | 2 | an error |
| `FORMAT('%t', 'x')` (GoogleSQL) | NULL | `x` |

## The fix

Each call is rewritten to a helper named for its dialect, which is the pattern the package already uses for LPAD, SUBSTR, CAST, MOD and division; the shared helpers keep the meaning of the dialect they were written for. MySQL's REGEXP_REPLACE gains its position, occurrence and match_type arguments; GoogleSQL's FORMAT gains %t and %T and hands every other verb to the formatting the printf path already did. The cast change splits the string path by dialect rather than rounding a float for all three. No public symbol is added or changed.

## Verification

`make test`, `make lint` and `make examples` pass. The whole differential sweep was re-run after the change and every one of the rows above now matches its engine. sqly builds and its tests pass against this checkout, and the queries its atago dialect specs pin answer exactly what they answered before.

The package had no benchmark, so this adds two. A query holding five of the newly rewritten calls translates in 13.4 us against 10.4 us before, allocating 64.6 KB against 49.8 KB: those calls are now rewritten and given the alias that keeps their result column name, at the same per-call cost as every other rewrite rule here, and a translation runs once per prepared statement rather than once per row. The per-row helpers move less, the largest being GoogleSQL's FORMAT at 1.31 times the printf call it replaces, which is the cost of printing a value where the old path answered NULL.

Closes #568
Closes #569
Closes #570
Closes #571
Closes #572
Closes #573
Closes #574

## Summary by CodeRabbit

* **New Features**
  * Improved MySQL, PostgreSQL, and GoogleSQL compatibility for logging, formatting, hexadecimal conversion, regular-expression replacement, and string slicing functions.
  * Added dialect-specific handling for integer casts, including truncation, overflow, exponent notation, and invalid input behavior.
  * Added GoogleSQL `%t` and `%T` formatting support.

* **Bug Fixes**
  * Corrected negative-length `LEFT` and `RIGHT` behavior.
  * Improved dialect-specific handling of unsupported function signatures and replacement flags.

* **Documentation**
  * Expanded documentation covering supported dialect-specific function behavior.


nao1215:
Applied the nitpick: the FORMAT locale case now sits above the M-23 comment, with a comment of its own, so the hexadecimal-literal note stays next to the cases it describes.

PR #631, opened 2026-08-26
fix(save): return from Close, check every overwrite target, and follow symlinks

### LINKED ISSUE #628: db.Close() never returns when auto-save is on and a transaction is still open
Found on main (2c06719) while running state-transition probes over auto-save.

## Phenomenon

A database opened with EnableAutoSave or EnableAutoSaveOnCommit never returns from db.Close() when a write transaction is still open, and the block is permanent: no busy timeout expires, no context cancels it, and the goroutine calling Close stays parked for the life of the process. The same program without auto-save returns from Close immediately, so this is a hang that filesql introduces on top of database/sql. It is easy to reach by accident because the common shape is a deferred Close plus an error path that returns without rolling the transaction back: what would otherwise be a leaked connection becomes a process that never exits.

## Reproduction

```go
package main_test

import (
	"context"
	"os"
	"path/filepath"
	"testing"
	"time"

	"github.com/nao1215/filesql"
)

func TestCloseHangsWithOpenTransaction(t *testing.T) {
	ctx := context.Background()
	dir := t.TempDir()
	path := filepath.Join(dir, "users.csv")
	if err := os.WriteFile(path, []byte("id,name\n1,alice\n"), 0o600); err != nil {
		t.Fatal(err)
	}

	b, err := filesql.NewBuilder().AddPath(path).EnableAutoSave("").Build(ctx)
	if err != nil {
		t.Fatal(err)
	}
	db, err := b.Open(ctx)
	if err != nil {
		t.Fatal(err)
	}

	tx, err := db.Begin()
	if err != nil {
		t.Fatal(err)
	}
	if _, err := tx.Exec("INSERT INTO users VALUES (2,'bob')"); err != nil {
		t.Fatal(err)
	}
	// The caller neither commits nor rolls back, then closes the database.

	done := make(chan error, 1)
	go func() { done <- db.Close() }()
	select {
	case err := <-done:
		t.Logf("Close returned: %v", err)
	case <-time.After(15 * time.Second):
		t.Fatal("db.Close did not return within 15s")
	}
}
```

```
=== RUN   TestCloseHangsWithOpenTransaction
    hang_test.go:47: db.Close did not return within 15s
--- FAIL: TestCloseHangsWithOpenTransaction (15.00s)
```

Dropping EnableAutoSave("") from the same test makes Close return nil in under a millisecond. The stack of the parked goroutine is the save reading the database:

```
modernc.org/sqlite.(*conn).retry(...)
modernc.org/sqlite.(*conn).step(...)
modernc.org/sqlite.(*stmt).query(...)
database/sql.(*DB).QueryContext(...)
```

The exact difference: Close blocks forever instead of returning, with or without a saved file.

## Expected behavior

db.Close() has to return. The database was closed while a transaction was still in flight, so the contents are not a state the caller has committed to, and writing them out is not what the caller asked for either way: the save should be skipped and Close should return an error naming the reason, so the caller learns that their transaction leaked and that nothing was written. EnableAutoSave documents that "The save runs once, when Close returns, so a database with auto-save is as safe to share across goroutines as one without it" (builder.go), and a Close that never returns is the opposite of that promise.

## Cause (guess)

This is a guess. autoSaveConnector.Close in save.go runs the save through the anchor connection, and the save reads every table with plain SELECTs against the shared-cache in-memory database. An uncommitted INSERT holds a write lock on that table, so the anchor's read gets SQLITE_LOCKED_SHAREDCACHE. modernc.org/sqlite handles that code in (*conn).retry by calling sqlite3_unlock_notify and blocking on a mutex with no deadline and no context, so the read waits until the transaction ends -- which never happens, because the only goroutine that could end it is the one blocked in Close. Nothing can be done about the waiting itself without changing that dependency, and no upstream report is being filed for it; what filesql can do is not start a read it knows will block. The connector already wraps every transaction as autoSaveTransaction, so it can count the transactions that are open and refuse to save when the count is not zero. Note that a caller who runs "BEGIN" through Exec rather than through database/sql bypasses that counting, which is worth saying in the doc comment rather than trying to detect.

## Tests to add

- In autosave_test.go: EnableAutoSave("") plus an INSERT inside a transaction that is never committed, then db.Close() run in a goroutine with a select on a timer. Close must return within a second or two with an error that mentions the open transaction, and the source file must still hold the original rows.
- In autosave_test.go: the same shape with EnableAutoSaveOnCommit(""), so the sibling timing does not keep the hang.
- In autosave_test.go: a read-only transaction left open at Close. That one does not take a write lock and does not block today, so pin whichever behavior the fix chooses rather than letting it drift.
- In autosave_test.go: the case that must keep working -- an unclosed *sql.Rows at Close. It holds a pooled connection but no transaction, and the save currently succeeds; a fix that keys on connections in use rather than on transactions would silently break it.
- In autosave_test.go: commit the transaction and then Close, asserting the file holds the committed row. This is the control that proves the refusal is about the open transaction and not about transactions in general.

### LINKED ISSUE #629: Overwrite auto-save replaces some files before failing on a source it can never write
Found on main (2c06719) while running state-transition probes over auto-save.

## Phenomenon

Auto-save in overwrite mode writes source files one at a time and stops at the first one it cannot write, so a set that contains a JSON or JSONL source is left half updated: the files that came earlier in the list hold the session's changes and the rest hold the old rows, and the caller sees only an error from Close. The failure is decided by the file's extension alone, so it is knowable before the caller has done any work at all, yet it is reported at the end of the session, after some files on disk have already been replaced. A caller who reloads that directory afterwards gets a mixture of two different points in time with nothing recording which files are which.

## Reproduction

```go
package main_test

import (
	"context"
	"os"
	"path/filepath"
	"testing"

	"github.com/nao1215/filesql"
)

func TestOverwriteStopsHalfway(t *testing.T) {
	ctx := context.Background()
	dir := t.TempDir()
	csvPath := filepath.Join(dir, "aaa.csv")
	jsonPath := filepath.Join(dir, "zzz.json")
	if err := os.WriteFile(csvPath, []byte("id,name\n1,alice\n"), 0o600); err != nil {
		t.Fatal(err)
	}
	if err := os.WriteFile(jsonPath, []byte(`[{"id":1,"name":"zed"}]`), 0o600); err != nil {
		t.Fatal(err)
	}

	b, err := filesql.NewBuilder().AddPath(csvPath).AddPath(jsonPath).EnableAutoSave("").Build(ctx)
	if err != nil {
		t.Fatal(err)
	}
	db, err := b.Open(ctx)
	if err != nil {
		t.Fatal(err)
	}
	if _, err := db.Exec("INSERT INTO aaa VALUES (2,'bob')"); err != nil {
		t.Fatal(err)
	}
	t.Logf("Close err: %v", db.Close())

	got, err := os.ReadFile(csvPath)
	if err != nil {
		t.Fatal(err)
	}
	t.Logf("aaa.csv after the failed save: %q", string(got))
}
```

```
=== RUN   TestOverwriteStopsHalfway
    half_test.go:39: Close err: auto-save failed: filesql: unsupported file format: /tmp/.../zzz.json cannot be written back because this package reads that format but does not write it; save to a directory with a format it writes instead
    half_test.go:45: aaa.csv after the failed save: "id,name\n1,alice\n2,bob\n"
--- PASS: TestOverwriteStopsHalfway (0.00s)
```

The exact difference: aaa.csv was replaced even though the save as a whole failed, and the caller was told about the JSON source only after that replacement had happened.

## Expected behavior

A save that cannot possibly succeed should not write any file, and the caller should learn about it before the session rather than after. Since overwriteFormatFor decides purely from the path, Build already has everything it needs: EnableAutoSave("") combined with a source in a format this package reads but does not write should fail at Build, the way an unusable output directory and an auto-save-plus-dialect combination already fail there, so the caller can fix the configuration before doing an hour of work that is about to be discarded. The save itself should also check the whole list before replacing the first file, so the same class of mistake cannot come back through a path Build does not see. Note this is a behavior change for a Build that succeeds today, and it belongs under Breaking Changes in the CHANGELOG.

## Cause (guess)

This is a guess. overwriteOriginalFiles in save.go loops over c.originalPaths and returns on the first error from overwriteOriginalFile, with no pass over the list beforehand; overwriteFormatFor is what rejects JSON and JSONL, and it only ever runs once the loop has reached that path. Build validates the auto-save output directory through validator.validateAutoSaveConfig but never looks at whether the collected paths can be written back, so overwrite mode is accepted for a set it can never save.

## Expected behavior of the sibling case

An output directory (export mode) is unaffected: it writes what DumpOptions says, so a JSON source is read and written out as CSV, and no source file is replaced. Only the empty-outputDir overwrite mode is at issue.

## Tests to add

- In autosave_test.go: EnableAutoSave("") with a .json source alongside a .csv source must fail at Build, with an error naming the JSON file. Assert the CSV is untouched, since Build never opens a database.
- In autosave_test.go: the same for .jsonl, so the sibling extension is covered rather than only the one that was found.
- In autosave_test.go: EnableAutoSave("./out") with the same two sources must still succeed and write both tables into the directory. This is what keeps the fix from over-rejecting.
- In save_test.go: a direct test of the pre-flight over a list of paths, covering a list that is entirely writable, one whose last entry is not, and one whose first entry is not, asserting that no file was replaced in the failing cases.

### LINKED ISSUE #630: Writing to a symlinked path replaces the link and leaves its target unchanged
Found on main (2c06719) while running state-transition probes over auto-save.

## Phenomenon

Writing to a path that is a symbolic link replaces the link with a regular file and leaves the file it pointed at untouched, so the caller's edit lands somewhere they never named and the data file they meant to update still holds the old rows. It happens on both write paths: EnableAutoSave("") writing a source back over itself, and DumpDatabase writing into a directory where the destination name already exists as a link. Symlinked data files are ordinary in deployment layouts and dotfile setups, where a working directory holds links into a shared or versioned location, and there the symptom is the worst kind: the save reports success, the tool that reads the real file sees no change, and the link that made the layout work is gone.

## Reproduction

```go
package main_test

import (
	"context"
	"os"
	"path/filepath"
	"testing"

	"github.com/nao1215/filesql"
)

func TestOverwriteReplacesSymlink(t *testing.T) {
	ctx := context.Background()
	dir := t.TempDir()
	real := filepath.Join(dir, "real.csv")
	if err := os.WriteFile(real, []byte("id,name\n1,alice\n"), 0o600); err != nil {
		t.Fatal(err)
	}
	link := filepath.Join(dir, "users.csv")
	if err := os.Symlink(real, link); err != nil {
		t.Skipf("symlink unsupported: %v", err)
	}

	b, err := filesql.NewBuilder().AddPath(link).EnableAutoSave("").Build(ctx)
	if err != nil {
		t.Fatal(err)
	}
	db, err := b.Open(ctx)
	if err != nil {
		t.Fatal(err)
	}
	if _, err := db.Exec("INSERT INTO users VALUES (2,'bob')"); err != nil {
		t.Fatal(err)
	}
	if err := db.Close(); err != nil {
		t.Fatal(err)
	}

	fi, err := os.Lstat(link)
	if err != nil {
		t.Fatal(err)
	}
	t.Logf("users.csv is still a symlink: %v", fi.Mode()&os.ModeSymlink != 0)
	linkBody, _ := os.ReadFile(link)
	realBody, _ := os.ReadFile(real)
	t.Logf("users.csv: %q", string(linkBody))
	t.Logf("real.csv:  %q", string(realBody))
}
```

```
=== RUN   TestOverwriteReplacesSymlink
    symlink_test.go:52: users.csv is still a symlink: false
    symlink_test.go:55: users.csv: "id,name\n1,alice\n2,bob\n"
    symlink_test.go:56: real.csv:  "id,name\n1,alice\n"
--- PASS: TestOverwriteReplacesSymlink (0.00s)
```

The same happens through DumpDatabase: with out/users.csv a link to target.csv, DumpDatabase(db, out) returns nil, out/users.csv becomes a regular file, and target.csv still reads "old\n".

The exact difference: real.csv should hold the new row and users.csv should still be a symbolic link; instead the row went into a new regular file at the link's own path and the link is gone.

## Expected behavior

A write to a symbolic link should follow it and replace the file it names, leaving the link in place, which is what every other program that edits a file in place does and what the caller means by "write it back to the file I loaded". The staging should therefore happen next to the resolved target, so the rename stays on the target's filesystem, and the resolved path is what gets replaced. A dangling link, one whose target does not exist, has no file to follow and should keep today's behavior of creating the file at the link's own path. Hard links are a different case and should stay as they are: a rename cannot preserve a link count, every atomic writer breaks them, and there is no path to follow the way there is with a symlink -- that is worth a sentence in the doc comment rather than a fix.

## Cause (guess)

This is a guess. atomicwrite.Write in internal/atomicwrite stages the new contents with createTempBeside(filepath.Dir(dest), ...) and finishes with commitStagedFile(tmpName, dest), and neither step looks at whether dest is a link. os.Rename replaces the link itself rather than following it, so the staged regular file takes the link's place. os.Stat(dest), which the same function uses to carry the destination's permissions over, does follow the link, so the new regular file inherits the target's mode -- which is why the result looks right until someone reads the real file. The fix belongs in that one function: resolve dest with filepath.EvalSymlinks (or os.Lstat plus os.Readlink, resolved against dest's directory) when it exists, and use the resolved path both for the staging directory and for the rename. Both callers, writeFileAtomically in atomic_write.go and the frame package's ToCSV and ToTSV, then get it at once.

## Tests to add

- In internal/atomicwrite's test file: writing through a symlink must leave the link a link and put the contents in its target. Cover a link pointing into a sibling directory as well as one in the same directory, since the staging directory changes with the target and a same-directory link would pass even if the resolution were only half done.
- In internal/atomicwrite's test file: a dangling symlink must still produce a file at the link's own path, and a plain regular destination must keep behaving exactly as it does now.
- In internal/atomicwrite's test file: a chain of two links must resolve to the final target.
- In autosave_test.go: the reproduction above, asserting through the real file rather than through the link, so the auto-save path is pinned and not only the helper.
- In dump_test.go: DumpDatabase into a directory whose destination name is a symlink, asserting the target received the rows and the link survived.
- All of these need a skip on the platforms where os.Symlink fails, since Windows needs a privilege for it; the existing tests that create symlinks show the shape to copy.

nao1215:
Three defects found by running state-transition probes over auto-save and the write-back path. All three are about a save that is reported as something other than what happened.

## What is fixed

`db.Close()` never returned when auto-save was on and a transaction was still open (#628). The save reads every table through the connector's own connection, an uncommitted write holds the lock on the table it touched, and the driver waits for that lock through `sqlite3_unlock_notify`, which has no deadline and answers to no context. The only goroutine that could release the lock was the one inside `Close`, so the wait never ended: no busy timeout expired, no cancellation reached it, and the same program without auto-save returned immediately. The shape that reaches it is an everyday one, a deferred `Close` and an error path that returns without rolling back, so a leaked connection became a hung process. The connector now counts the transactions the caller has begun and not yet finished, and a close with any of them still open skips the save and reports that it did.

Overwrite auto-save replaced some files before failing on a source it could never write (#629). It wrote the sources one at a time and stopped at the first refusal, so a set holding a JSON, a JSONL or a bzip2 source came out of a failed save with its earlier files carrying the session's rows and the rest carrying the old ones, with nothing on disk saying which was which. Whether a source can be written back is decided by its name alone, so `Build` now answers it before the session, and the save keeps the same check ahead of its loop.

A write to a path that is a symbolic link replaced the link and left its target unchanged (#630). The staged file was renamed onto the link itself, so an overwrite auto-save and a `DumpDatabase` into a directory of links both reported success over a data file that never moved. The staging and the rename now both move to the file the link points at, which also keeps the rename inside the target's filesystem.

## Notes on the choices

A transaction that only read is treated the same way as one that wrote. It never hung, since reading takes no write lock, but one rule is easier to rely on than a rule that depends on what the transaction happened to run. An unclosed `*sql.Rows` still saves, and there is a test for it: it holds a pooled connection but no transaction, so a fix that keyed on connections in use rather than on transactions would have broken it silently. A transaction begun by running `BEGIN` through `Exec` is invisible to the count and a close during one still waits forever, which is said in the doc comment rather than guessed at from the SQL.

`Build` refusing a source overwrite mode could never write back is a breaking change, and the CHANGELOG carries the migration: pass an output directory. The save failed on those sources either way; what changes is that it fails before the session rather than at the end of it.

Hard links are deliberately left alone. A rename cannot preserve a link count, every atomic writer breaks them, and there is no path to follow the way there is with a symlink.

## Verification

`make test`, `make lint` and `go test -race ./...` are clean, and the branch cross-builds for windows/amd64 and darwin/arm64. Every function added is at 100% statement coverage and the total is 93.7%. `benchstat` over `BenchmarkOpenWithAutoSave`, `BenchmarkOverwriteWorkbook` and `BenchmarkDumpText` with `-count=5` shows +3.25% geomean on time and no change at all in bytes or allocations per operation, which is the one extra path resolution a write now does. sqly builds and its tests pass against this checkout with a local replace, which was then dropped.

Closes #628
Closes #629
Closes #630

## Summary by CodeRabbit

- **Bug Fixes**
  - Auto-save skips saving when transactions remain open and reports the skipped operation.
  - Failed commits no longer leave transactions marked as open.
  - Overwrite saves validate all targets before replacing files, preventing partial updates.
  - Symbolic-link writes update their targets while preserving the links.
  - Unsupported formats and compression codecs are rejected earlier with clearer errors.
- **Documentation**
  - Clarified auto-save and database dump behavior.
- **Breaking Changes**
  - Overwrite auto-save requires writable source formats unless an output directory is specified.


nao1215:
The merge risk is real and is fixed in 3b2db0c. A commit that returned an error left the transaction in the connector's count forever, because database/sql calls neither Commit again nor Rollback after that, so every later close refused a save it should have run. The count is now dropped whether the commit succeeded or not, which is safe here: modernc.org/sqlite forces a rollback itself when a commit fails, so the connection holds no lock for the save to wait on. TestAutoSaveTransaction_CommitThatFailedStopsCountingAsOpen pins it.

PR #632, opened 2026-08-26
fix(dialect): read the clock in UTC and fix it for the statement

### LINKED ISSUE #622: The clock functions read local time and re-read the clock per row
Raised in review of the GoogleSQL work, against every dialect rather than one.

## Phenomenon

The clock functions read the host's local time zone and re-read the clock on every call, and neither is what any of the three engines does. `NOW()`, `CURDATE()`, `CURTIME()`, `UNIX_TIMESTAMP()`, PostgreSQL's `CLOCK_TIMESTAMP` and its siblings, and GoogleSQL's `CURRENT_DATETIME` all call Go's `time.Now()` and format it in whatever zone the process is running in. BigQuery's `CURRENT_DATETIME()` is UTC unless a zone is named; MySQL's `NOW()` and PostgreSQL's `now()` use the session's zone, which filesql has no equivalent of. So the same query answers a different hour depending on the machine it runs on, and two callers comparing results across machines see a difference the data does not have.

Separately, each of these engines fixes the reading once per statement -- MySQL and PostgreSQL at statement start, BigQuery likewise -- while a SQLite user-defined function registered as non-deterministic is called once per row. `SELECT NOW() FROM t` over a large table can therefore answer several different times in one result, and a row can be timestamped after the row below it.

SQLite's own `CURRENT_TIMESTAMP` has neither problem: it is UTC and fixed per statement. So a query mixing `NOW()` with `CURRENT_TIMESTAMP` compares two different clocks.

## Reproduction

```go
package main

import (
	"database/sql"
	"fmt"

	_ "modernc.org/sqlite"
)

func main() {
	defer db.Close()
	db.Exec(`CREATE TABLE t (n INTEGER); INSERT INTO t VALUES (1),(2),(3)`)

	q, err := dialect.Translate(dialect.MySQL, `SELECT NOW(), CURRENT_TIMESTAMP FROM t`)
	if err != nil {
		panic(err)
	}
	rows, err := db.Query(q)
	if err != nil {
		panic(err)
	}
	defer rows.Close()
	for rows.Next() {
		var now, current string
		if err := rows.Scan(&now, &current); err != nil {
			panic(err)
		}
		fmt.Printf("NOW()=%s CURRENT_TIMESTAMP=%s\n", now, current)
	}
}
```

The two columns differ by the machine's offset from UTC. Run it with TZ set to something other than UTC and the difference is that offset; run the same query over enough rows and the first column can change within one result while the second cannot.

## Expected behavior

This is one decision for every clock function rather than a per-dialect one, which is why it is its own issue rather than part of the dialect work. The two questions are which zone and which instant.

For the zone, filesql carries none: no column has one, no cast produces one, and the dialect package refuses the arguments that name one. Reading the clock in UTC would make the answer the same on every machine and would match SQLite's own `CURRENT_TIMESTAMP` and BigQuery's default; it would differ from MySQL and PostgreSQL, whose answer depends on a session zone that has no counterpart here. Reading it locally, which is what happens today, matches nothing exactly and varies by machine.

For the instant, a value fixed at the start of the statement is what all three engines answer, and it is what SQLite's own `CURRENT_TIMESTAMP` answers. Getting it needs something the current registration has no access to: a scalar function sees no statement boundary. The shape that would work is a value captured per connection or per statement and read by the helpers, which is a change to how they are registered rather than to what they compute.

Either change moves an answer that callers may be relying on, so it belongs in a release that says so.

## Cause (guess)

A guess: `fnNow`, `fnCurdate` and `fnCurtime` in `dialect/functions.go` were written as one-line `time.Now().Format(...)` calls, and every clock function added since has followed them, including the four added for PostgreSQL and the one for GoogleSQL. Nothing decided on local time; it is Go's default for a formatted `time.Time`.

## Tests to add

- A test that sets a non-UTC zone for the process and asserts the answer is the same as under UTC, which is what pins the zone decision whichever way it goes.
- `SELECT NOW() FROM t` over enough rows to cross a second boundary answers the same value in every row, which pins the statement-scoping decision.
- `NOW()` and SQLite's own `CURRENT_TIMESTAMP` in one query agree, which is the property that makes the two readings comparable.
- The same three for each of the clock functions in all three dialects, since they should not diverge from each other again.

nao1215:
Clears the older of the two open bug issues, and writes down one invariant a differential sweep found holding.

## The clock

The clock helpers called `time.Now()` and formatted it in whatever zone the process was running in, so `NOW()` answered a different hour on every machine and disagreed with SQLite's own `CURRENT_TIMESTAMP` in the same row. This package carries no time zone -- no column has one, no cast produces one, and a zone argument is refused -- so UTC is the only reading that is the same everywhere, and it is what SQLite's clock and BigQuery's default already answer.

They were also registered as non-deterministic, which makes SQLite call them once per row, so `SELECT NOW() FROM t` over a large table answered several times in one result and a row could be stamped after the row below it. Every engine this package translates fixes the reading when the statement starts. `NOW`, `CURDATE`, `CURTIME`, `UNIX_TIMESTAMP`, PostgreSQL's `now`, `statement_timestamp` and `transaction_timestamp`, and BigQuery's `CURRENT_DATETIME` are now registered as deterministic, which is what fixes them: SQLite computes such a function once per execution and computes it again for the next statement. Both halves of that were measured before the change rather than assumed -- a no-argument deterministic helper was called once and answered one value across 300 rows, where a non-deterministic one was called 300 times, and a prepared statement re-executed three times got a new value each time.

`clock_timestamp` and `timeofday` stay per-row. Advancing during a statement is the whole of what separates them from `now` in PostgreSQL, so filesql now makes a distinction it used to collapse. `RAND`, `GENERATE_UUID` and `GEN_RANDOM_UUID` are unaffected. `TO_TIMESTAMP` was in the non-deterministic table although it reads no clock, and has moved across with the pure functions.

Both changes move an answer a caller may have been relying on, and the CHANGELOG says so.

## The regression test needs a second process

A test host already on UTC cannot tell a local reading from a UTC one, and CI runs on exactly such machines, so the zone assertions run in a child process with `TZ` set to `Asia/Tokyo`. Go reads the zone once, so setting it from inside the test is too late. The child skips itself where the platform does not take the zone from `TZ`.

The per-statement half cannot be asserted from SQL at all: a second is too coarse, and every row of a fast statement lands in the same second whichever way the helper is registered. The registration table is the assertion instead, which is exact and does not flake.

## What the digging found, and did not

The session ran a differential sweep before reaching for this issue, and it came back clean, which is worth recording so it is not repeated: chunk sizes 1, 2, 3, 7, the default and 100000 over 60 random CSVs and 125 random files in five formats built identical tables; a path, an `fs.FS` entry, a reader and `LoadInto` agreed on the same file; type widening across a chunk boundary agreed for seven different TEXT-forcing values at six positions; a dump-and-reload round trip over 35 adversarial values in five output formats preserved every one, with the three refusals being the documented TSV and LTSV ones; and `FuzzParseFormats` ran 13.4 million executions without a crash. The only invariant among those with no test of its own was route agreement, so this adds one.

Closes #622

## Summary by CodeRabbit

- **New Features**
  - Clock functions now consistently use UTC.
  - Most clock functions return one stable value throughout each statement, while changing-clock and random functions continue updating per row.
  - GoogleSQL and PostgreSQL timestamp functions now follow consistent UTC behavior.

- **Bug Fixes**
  - Improved consistency when loading databases from paths, filesystems, readers, or caller-provided sources, including schema, type inference, empty values, and CSV quoting.


nao1215:
The merge risk is real and is fixed in 5c4b93f. Registering the fixed clock as deterministic takes it off the row but not quite onto the statement: SQLite folds each occurrence of the call separately, so `SELECT NOW(), NOW()` read the clock twice and the two answers differed whenever those microseconds straddled a second. Measured rather than assumed: a counting deterministic helper is called twice for `SELECT probe(), probe() FROM t` over 50 rows, and twice again for the UNION ALL and the correlated-subquery shapes. The fixed clock now shares one reading for a millisecond, which is wider than the gap between occurrences and far narrower than the second these functions are formatted to; a statement beginning inside that window sees the previous reading, which is at most a millisecond old and never earlier than what an earlier statement answered. clock_timestamp reads the clock itself, since moving while a statement runs is what separates it from now. TestClockFunctionsAreFixedPerStatement gains an occurrence case and TestClockSharesOneReadingWithinItsWindow pins the mechanism, which SQL cannot show because the answers are whole seconds.

PR #636, opened 2026-08-26
test(codec): budget the fuzz target by the ceilings this package enforces

### LINKED ISSUE #635: FuzzCodecReader fails on a zstd frame the package accepts, so the nightly job goes red at random
Found by the nightly Fuzz workflow, which went red on run 33005872606 and has been green on every run before it.

## Phenomenon

FuzzCodecReader fails on a zstd frame the package deliberately accepts, and it fails only sometimes, so the nightly job goes red on a schedule nobody can predict and for no defect. The target budgets what opening a stream may allocate at 64 MiB, while the package's own ceiling lets a zstd frame declare a window of up to 128 MiB and an xz stream a dictionary of up to 256 MiB. README states that ceiling as the intended cost -- "A damaged file therefore costs a fixed ceiling rather than whatever its header names" -- so a 29-byte frame declaring a 64 MiB window is exactly what the package promises to spend, and the test calls it a failure. A red nightly that is not a defect is worse than no nightly: the next real finding is read as noise.

The second half of why it is intermittent is the measurement. runtime.ReadMemStats reports what the whole process has allocated, and klauspost/compress starts decoder goroutines that allocate the window buffer after zstd.NewReader has returned. Whether that lands inside the two ReadMemStats calls depends on the scheduler, so the same input fails on a loaded four-core runner and passes locally: five minutes of local fuzzing over the same corpus does not reproduce it, and re-running the saved input passes.

## Reproduction

The input the run saved, which is a valid zstd frame header declaring a 64 MiB window (window descriptor 0x80: exponent 16, so 2^26) followed by bytes that are not compressed data:

```
[]byte("(\xb5/\xfd\x04\x80\x81\x00\x00id:name\n1,alice\n\x8c\xd7\xcf\xef")
```

What it costs, measured over whole open-read-close cycles so the decoder goroutines cannot hide behind the measurement:

```go
func TestZstdCycleCost(t *testing.T) {
	data := []byte("(\xb5/\xfd\x04\x80\x81\x00\x00id:name\n1,alice\n\x8c\xd7\xcf\xef")
	const n = 200
	var before, after runtime.MemStats
	runtime.ReadMemStats(&before)
	for range n {
		dec, err := zstd.NewReader(bytes.NewReader(data), zstd.WithDecoderMaxWindow(128<<20))
		if err != nil {
			continue
		}
		_, _ = io.Copy(io.Discard, io.LimitReader(dec, 1<<20))
		dec.Close()
	}
	time.Sleep(200 * time.Millisecond)
	runtime.ReadMemStats(&after)
	total := after.TotalAlloc - before.TotalAlloc
	fmt.Printf("%d cycles allocated %d MiB, %d KiB per cycle\n", n, total>>20, (total/n)>>10)
}
```

```
200 cycles allocated 13002 MiB, 66570 KiB per cycle
```

A frame the library itself wrote costs 8 KiB per cycle at every max-window setting from 1 MiB to 128 MiB, so the 65 MiB belongs to this frame's declared window and not to zstd in general:

```
maxWindow=   1 MiB  per cycle       8 KiB  decoded=60 failed=0
maxWindow=  64 MiB  per cycle       8 KiB  decoded=60 failed=0
maxWindow= 128 MiB  per cycle       8 KiB  decoded=60 failed=0
```

The exact difference: the target's budget is 64 MiB, the package's zstd ceiling is 128 MiB and its xz ceiling is 256 MiB, and a stream may legitimately spend up to those.

## Expected behavior

The budget has to be the same number the package enforces, or the target is testing a rule the package does not have. It should be derived from maxZstdWindow and maxXZDictionary rather than written independently, so raising or lowering a ceiling moves the test with it and the two cannot drift apart again. That keeps everything the target was written for: the two bugs it guards against reached 4 GiB from a 28-byte xz stream and 512 MiB from a 14-byte zstd stream, both far above 256 MiB, and a third codec allocating without a ceiling would still be caught.

The frame above should become a seed, so the case of a stream that is accepted and expensive runs on every test rather than only when the fuzzer rediscovers it.

## Cause (guess)

This is a guess. maxSetupAlloc in internal/codec/codec_fuzz_test.go is a const of its own, chosen as "wide enough for a decompressor's own buffers and far below the sizes the two fixed bugs reached". It was written before the ceilings were what they are now, or without noticing that 64 MiB sits under both of them; either way nothing ties it to the constants in codec.go, so the two say different things about the same rule.

## Tests to add

- Derive the budget from the ceilings in codec.go and add a test asserting the budget is above both, so a later change to either ceiling that would make the target fail on an accepted stream is caught in the same package rather than by a nightly.
- Seed the target with the frame above, and with a zstd frame declaring the largest window the package accepts, so the most expensive accepted stream is exercised every run.
- Seed it with an xz stream declaring a dictionary just under and just over maxXZDictionary, which is the sibling ceiling and has no seed of its own today.

nao1215:
The nightly Fuzz workflow went red for the first time, on a stream this package is documented to accept.

## What was wrong

`FuzzCodecReader` budgets what opening a stream may allocate before any of it is read, and that budget was a number of its own: 64 MiB. It sits below both ceilings the package enforces -- a zstd frame may declare a 128 MiB window and an xz stream a 256 MiB dictionary -- and README states that ceiling as the intended price: "A damaged file therefore costs a fixed ceiling rather than whatever its header names". So the 29-byte frame the fuzzer found, which declares a 64 MiB window through descriptor 0x80, is exactly what the package promises to spend, and the target called it a failure.

It failed only sometimes. `runtime.ReadMemStats` reports what the whole process has allocated, and klauspost/compress allocates the window in decoder goroutines that outlive `zstd.NewReader`, so whether the allocation lands between the two readings depends on the scheduler. Five minutes of local fuzzing over the same corpus does not reproduce it and re-running the saved input passes, while a loaded four-core runner hit it. A nightly that goes red at random and for no defect is worse than no nightly: the next real finding reads as noise.

## What the numbers are

Measured over whole open-read-close cycles, so the goroutines cannot hide behind the measurement:

```
200 cycles of the found frame allocated 13002 MiB, 66570 KiB per cycle
```

A frame the library itself wrote costs 8 KiB per cycle at every max-window setting from 1 MiB to 128 MiB, so the 65 MiB belongs to this frame's declared window and not to zstd in general. The window it declares is 2^26, which is 64 MiB, under the package's 128 MiB ceiling and therefore accepted.

## What changed

The budget is derived from `maxXZDictionary` and `maxZstdWindow` rather than written independently, so raising or lowering a ceiling moves the test with it. A test pins it between the two numbers that give it meaning: above every declaration the package accepts, so it cannot fail on a stream that is merely expensive, and below the smaller of the two allocations the target exists to catch, which were 4 GiB from a 28-byte xz stream and 512 MiB from a 14-byte zstd one. A ceiling raised past that range now has to be argued for in this package rather than discovered by a nightly.

The frame the fuzzer found is a seed, so the most expensive stream the package accepts is opened on every run. The xz ceiling had no seed of its own, so streams on either side of it are seeded too: property 32 encodes exactly `maxXZDictionary` and is accepted, 34 encodes twice that and is refused.

Nothing in the product changed. The ceilings are what README says they are and what the two earlier fixes set them to.

Closes #635

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed fuzzing allocation limits to align with accepted xz and zstd stream limits.
  - Improved detection of excessive memory allocation during codec processing.

- **Tests**
  - Added coverage for allocation budget boundaries.
  - Added representative xz and zstd streams, including oversized declarations, to fuzzing seeds.

- **Documentation**
  - Documented the codec fuzzing allocation fix in the changelog.


nao1215:
The zstd half of that gap is real and is closed in the commit just pushed. The xz ceiling already had its pair -- property 32 encodes exactly maxXZDictionary and is accepted, 34 encodes twice that and is refused -- and zstd now has the same: exponent 17 selects 2^27, which is exactly maxZstdWindow and is accepted, and 18 selects twice that and is refused. Both accepted seeds are the expensive case on purpose, since they are the most either format may ask this package to spend, and the budget sits above both. The windows were computed rather than assumed: exponent 11, 17, 18 and 19 give 2, 128, 256 and 512 MiB, and properties 22, 32, 34 and 40 give 8, 256, 512 and 4096 MiB.

PR #821, opened 2026-08-28
refactor(dialect): rebuild the translation as a compiler frontend

nao1215:
The dialect package tokenized a query and then rewrote the token slice in place, one pass per construct. Nothing between the lexer and the renderer knew what an expression was, so every pass that needed an operand rediscovered it by walking tokens outward and guessing where the operand began and ended; the package carried `primaryStartBack`, `primaryEndForward`, `operandChainStartBack`, `chainStartBack` and `adjacentCallEnd` for that, and each of them encoded a partial, private theory of SQL's grammar.

Three kinds of defect followed, and all three were paid for in bugs. Operator precedence lived in the passes rather than in one place, so the order they ran in became part of the language's meaning: MOD had to be rewritten before DIV, `!` before `^`, the typed-literal pass before the interval pass, and every new construct had to be inserted at the right point in a sequence whose constraints were recorded only in comments. An operand boundary found by scanning is a guess, and the guesses were wrong at the edges: `SOUNDS LIKE`, `LIKE ANY`, the PostgreSQL bit-string literal and the `#-` operator each produced SQL that could not parse. And anything no pass recognized was handed to SQLite unchanged, which made the supported language accidental: a construct worked whenever SQLite happened to accept the same text, with whatever meaning SQLite gives it, so `CREATE TABLE x LIKE y` was accepted and built a table with no columns.

## What this changes

A query now goes through a lexer, a parser that builds a typed syntax tree, a lowering layer that rewrites the tree into one SQLite can execute, and a renderer that writes the SQL back out. Expression parsing is a Pratt parser over one precedence table per dialect, which is the whole of the package's knowledge of what binds tighter than what; the table reads the dialect where the three disagree, as they do about XOR and about what `^` means. Statement parsing is recursive descent over SELECT, the set operations, the data statements and the three objects SQLite can create. The lowering rules are given a node whose children are already lowered, so a rule reads a tree rather than neighbouring tokens and no rule depends on another having run first. The renderer decides quoting and parenthesization from the tree, and errors on a node SQLite cannot express rather than writing text that will not parse.

The layers are separate packages under `dialect/internal`, with the dependency running one way: `token`, then `ast`, then `parser`, `lower` and `render`, with `runtime` -- the helper functions registered into the driver -- reachable only by name. `dialect/ARCHITECTURE.md` records why, for whoever changes this next. The public API is unchanged apart from one new sentinel.

## Fail closed

A construct outside the supported subset is refused with the line and column it was written at, rather than forwarded to SQLite. `ErrInvalidSyntax` means the query could not be read, `ErrUnsupportedSyntax` that SQLite cannot express the construct, and the new `ErrUnsupportedFeature` that this package does not model it. The subset is listed in the package documentation and is the contract: SELECT with every clause, WITH and WITH RECURSIVE, the set operations, the joins, the data statements, and the DDL SQLite has. Everything else -- `GRANT`, `LOCK TABLES`, `SHOW`, `CREATE SEQUENCE`, `CREATE MATERIALIZED VIEW`, `CREATE TABLE ... LIKE`, changing a column's type -- is refused by name.

Several of the open issues about statements reaching SQLite raw are fixed by that alone, and a few are fixed properly: `SERIAL` now becomes `INTEGER PRIMARY KEY AUTOINCREMENT` instead of a column nothing numbers, a MySQL column `COMMENT` is dropped instead of being folded into the declared type, and the storage clauses that describe a physical layout -- `ENGINE`, `CHARSET`, `OPTIONS`, `CLUSTER BY`, `UNLOGGED`, `CONCURRENTLY`, `CASCADE`, MySQL's index hints -- are read and dropped so the statement runs.

## Behaviour that changes

Comments are dropped from the output. The tree does not model them, SQLite acts on nothing in them, and the previous renderer had to defuse a nested comment's delimiters to keep its own output parseable. A query that holds only comments is refused for the same reason MySQL and PostgreSQL refuse it: there is no statement in it. sqly reads that case today by translating to an empty string and reporting its own message, so it needs a one-line follow-up to recognize the error instead; nothing else downstream is affected.

The translated SQL is also spelled the way a reader would spell it. An identifier keeps its bare form unless the caller quoted it or it collides with a SQLite keyword, where before every name was written in quotes; a sign stays against its operand; and the parentheses a rewrite added around every operand are gone. An alias the caller left out is written in, so MySQL's `SELECT a` with a backticked name beside it comes out as `SELECT a AS "b"` rather than as two tokens whose relationship the reader has to work out, and `LIMIT 5, 10` comes out as `LIMIT 10 OFFSET 5`.

## Measurement

Translation is faster and allocates a third of what it did, because the tree is built once where the passes copied the whole token slice on every one of them. Three runs each, on the package's own benchmark: MySQL 19.1us and 91.6 KB per translation before, 13.4us and 19.6 KB after; PostgreSQL 14.9us and 74.5 KB before, 11.1us and 16.5 KB after; GoogleSQL 12.8us and 48.0 KB before, 11.2us and 16.9 KB after.

## Tests

Every dialect regression test moved next to the layer that now implements what it covers and still passes: the value tables against mysql:8.4, postgres:17 and BigQuery are in `internal/runtime`, and the translation tables in `internal/lower`. Added on top: a precedence and associativity suite that asserts the shape of the tree rather than the text it renders to, a statement suite that walks the tree, an end-to-end suite over every statement form, a renderer suite that builds un-lowered nodes and holds the renderer to refusing them, a round-trip suite that renders and re-reads, and fuzz targets for the lexer and the parser beside the existing one for translation. `go test ./...`, `go test -race ./...`, `make lint`, a Windows and a macOS cross-build, and sqly's own suite all pass.

## Summary by CodeRabbit

* **Breaking Changes**
  * Dialect translation now rejects unsupported or invalid SQL with line and column diagnostics instead of forwarding it to SQLite.
  * Empty and comment-only queries are rejected, and comments are removed from translated SQL.

* **Changed**
  * Improved MySQL, PostgreSQL, and GoogleSQL compatibility for functions, operators, casts, aggregates, aliases, identifiers, and date/time handling.
  * MySQL `LIMIT 5, 10` is rendered as `LIMIT 10 OFFSET 5`.
  * Added `ErrUnsupportedFeature` for recognized but unimplemented SQL features.

* **Documentation**
  * Updated translation guidance and added an unsupported-feature example.


nao1215:
Thanks — eight of the nine findings were real, and each of them was a query that ran and answered something the caller had not asked for. All eight are fixed in 20c189f, with a regression test each and a CHANGELOG entry naming them.

Fixed:

- Set-operation grouping (`render/stmt.go`). SQLite has no precedence among the compound operators and reads them left to right, so `SELECT 1 UNION SELECT 2 INTERSECT SELECT 3` was regrouped. A set operation on the right of another is written as `SELECT * FROM (...)` now; on the left it needs nothing, since that is the order SQLite already takes.
- PostgreSQL's "any other operator" level (`parser/precedence.go`). `||`, `#`, the shifts, the bitwise operators and the JSON operators all sit below addition in PostgreSQL's grammar and each had a level of its own here. They share one level below `precAddSub` now, and BigQuery's `||` moved to the level it multiplies at, which is where its own documentation puts it.
- SQLite's precedence for `->` and `->>` (`render/expr.go`), which bind tighter than arithmetic there.
- `LIKE ... ESCAPE` (`lower/shared.go`). Leaving it on SQLite's own LIKE meant matching with SQLite's rules rather than the dialect's, so under PostgreSQL `'A' LIKE 'a' ESCAPE '!'` was true. Rather than refuse it, `like_sensitive` and `like_insensitive` take the escape character now; an empty ESCAPE turns escaping off, which is PostgreSQL's reading of it.
- `CONVERT(x USING ...)` (`lower/mysql.go`). `utf8CharsetName` accepted `binary`, `ucs2`, `utf16` and `utf32`, and the conversion was then dropped. Only the UTF-8 spellings are accepted now; the same list backs the `_charset'...'` introducer.
- `FILTER` and `DISTINCT` on an expanded aggregate (`lower/aggregate.go`). Both are refused by name, for the same reason `OVER` already was: the expansion is several aggregates inside arithmetic and the clause belongs to none of them.
- `FETCH FIRST ROW ONLY` (`parser/select.go`), which lost its limit and returned every row.
- The `ON CONFLICT` index predicate (`parser/stmt.go`), which picks the partial index the upsert resolves against and was parsed and thrown away.
- `ON DELETE SET NULL` and `ON DELETE SET DEFAULT` (`parser/ddl.go`), truncated after the `SET`, with the `NULL` then taken for a column constraint the caller never wrote. The referential actions are read by name now.
- The short `U&'\+41424'` escape (`parser/primary.go`), an off-by-one that read past the end of the string and panicked. A query is untrusted text, so it is a refusal.
- The two unbounded recursive paths (`parser/select.go`, `parser/stmt.go`). `parseQueryPrimary` and `parseExplain` carry the depth guard now, with fuzz seeds for both shapes.

Declining one:

- Mapping `_ci`, `_ai_ci` and `_as_ci` to `NOCASE` in `sqliteCollation`. You are right that NOCASE folds only ASCII where those collations fold Unicode, and the package documents that limit for the comparison operators already. It is not new here: the mapping is carried over unchanged from `sqliteCollationFor`, and it was added deliberately to replace the two worse outcomes a raw `COLLATE` produced -- SQLite raising about a collation the caller never wrote, or accepting the clause and applying nothing. Turning the approximation into a refusal would take away a clause that works today for ASCII data, which is most of what a CSV holds, and it is a decision about what the package promises rather than about this rewrite. If it should become a refusal I would rather do it in its own change with its own issue, so the removal is visible in the CHANGELOG for what it is.

One note on the aggregate finding: a renamed aggregate keeps both clauses, since the call it becomes is still one call; only the ones that expand into an expression refuse them. There is a test for that too.

nao1215:
Second round: all four fixed, each measured against postgres:17.10 or mysql:8.4.11 rather than reasoned about.

- `CONVERT(x USING utf8)` and `utf8mb3`. You are right that MySQL's `utf8` is `utf8mb3` and cannot carry a character outside the basic plane, so dropping the conversion is not the no-op it is for `utf8mb4`. Only `utf8mb4` and the empty name are accepted now, on both the `CONVERT ... USING` path and the `_charset'...'` introducer, with regressions for `utf8`, `utf8mb3` and `_utf8'abc'`.
- Surrogate pairs and the zero character in `U&'...'`. `U&'\D83D\DE00'` wrote each half on its own and produced two replacement characters. The halves are read together now; a half with no partner and `\0000` are refused. PostgreSQL answers `😀`, `invalid Unicode surrogate pair` and `invalid Unicode escape value` for those three, and so does this.
- A trailing escape in a LIKE pattern. Measured rather than assumed, because the two dialects differ and PostgreSQL's own rule is narrower than "the pattern ends with the escape": `'!' LIKE '!' ESCAPE '!'` raises, `'a!' LIKE 'a!' ESCAPE '!'` raises, and `'ab' LIKE 'ab!' ESCAPE '!'` answers false, because the walk runs out of subject before it reaches the escape. MySQL answers 1, 1 and 0 for the same three, reading a trailing escape as itself. Both readings are implemented, reported from inside the match where PostgreSQL reports it, and `likeCompare` takes the strictness as its own argument rather than deriving it from case sensitivity.
- `VALUE` in `clauseKeywords`. You are right that `SELECT * FROM t value` is valid PostgreSQL and the entry broke it. The map is unchanged again; instead an INSERT's target takes an alias only after `AS`, which is how every dialect writes one, so `INSERT INTO t VALUE (1)` reads and `SELECT * FROM t value` still does too.

One correction to the finding on the trailing escape: rejecting whenever the last rune equals the escape would have refused `'ab' LIKE 'ab!' ESCAPE '!'`, which PostgreSQL answers rather than refusing.

Issue #810, opened 2026-08-27
NULLIF, DATE_FORMAT and the REGEXP family answer a value where MySQL answers NULL or refuses


nao1215:
Found on main (e17864b) by running a second sweep of generated MySQL calls against a real mysql:8.4 container.

## Phenomenon

Three MySQL behaviours around an empty or unmatched value differ, each in a way that answers a plausible value rather than what MySQL answers.

`NULLIF('abc', 0)` is NULL in MySQL, because the comparison coerces the string to a number and 0 equals 0; here it answers 'abc'. `NULLIF('', 0)` is NULL there and the empty string here. `DATE_FORMAT(d, '')` is NULL in MySQL -- an empty format produces nothing, which MySQL reports as NULL rather than as the empty string -- and here it is the empty string. `REGEXP_LIKE(x, '')` is an error in MySQL, which refuses an empty pattern, and here it answers 1 for every subject.

## Reproduction

```go
package main

import (
	"database/sql"
	"fmt"

	"github.com/nao1215/filesql/dialect"
	_ "modernc.org/sqlite"
)

func main() {
	_ = dialect.RegisterFunctions()
	db, _ := sql.Open("sqlite", ":memory:")
	for _, expr := range []string{
		"NULLIF('abc', 0)", "NULLIF('', 0)", "NULLIF('abc', 'abc')",
		"DATE_FORMAT('2024-02-29', '')", "REGEXP_LIKE('abc', '')", "REGEXP_SUBSTR('abc', '')",
	} {
		q, _ := dialect.Translate(dialect.MySQL, "SELECT "+expr)
		var v any
		if err := db.QueryRow(q).Scan(&v); err != nil {
			fmt.Printf("%-32s ERR %v\n", expr, err)
			continue
		}
		fmt.Printf("%-32s = %#v\n", expr, v)
	}
}
```

```
NULLIF('abc', 0)                 = "abc"
NULLIF('', 0)                    = ""
NULLIF('abc', 'abc')             = <nil>
DATE_FORMAT('2024-02-29', '')    = ""
REGEXP_LIKE('abc', '')           = 1
REGEXP_SUBSTR('abc', '')         = ""
```

MySQL 8.4 answers NULL, NULL, NULL, NULL, and raises `Illegal argument to a regular expression` for the last two.

The differences are that NULLIF compares with SQLite's rules rather than MySQL's, that an empty format is the empty string rather than NULL, and that an empty pattern matches everything rather than being refused.

## Expected behavior

[...]

PR #852, opened 2026-08-31
fix: honor the transaction options, count a BEGIN statement, and compare NULLIF the way MySQL does

### LINKED ISSUE #849: BeginTx accepts a sql.TxOptions it does not implement: ReadOnly permits writes and every isolation level is taken
Found on main (544c281) with Go 1.26.4 on Linux, while walking the auto-save and transaction state transitions of a database this package opens.

## Phenomenon

Every database this package hands back accepts a `sql.TxOptions` it does not implement and then ignores it. `db.BeginTx(ctx, &sql.TxOptions{ReadOnly: true})` returns a transaction that happily runs an UPDATE, so a caller who asked for a read-only transaction as a guard rail gets no guard rail and writes into the in-memory database that auto-save will later write to the file. `db.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelReadUncommitted})` -- and every other level, up to `sql.LevelLinearizable` -- is accepted as well, though SQLite provides serializable transactions and nothing else. This holds for all four handle kinds: the plain one from `DBBuilder.Open`, the auto-save one, the dialect one, and `DBBuilder.OpenReadOnly` (whose read-only-ness comes from the `query_only` pragma on the handle, not from the transaction option).

## Reproduction

```go
package main

import (
	"context"
	"database/sql"
	"fmt"
	"os"
	"path/filepath"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
	dir, _ := os.MkdirTemp("", "tx")
	_ = os.WriteFile(filepath.Join(dir, "users.csv"), []byte("id,name\n1,alice\n"), 0o600)
	db, err := filesql.NewBuilder().AddPath(dir).Open(ctx)
	if err != nil {
		panic(err)
	}
	defer func() { _ = db.Close() }()

	tx, err := db.BeginTx(ctx, &sql.TxOptions{ReadOnly: true})
	fmt.Println("begin read-only:", err)
	_, err = tx.ExecContext(ctx, "UPDATE users SET name='written' WHERE id=1")
	fmt.Println("write inside the read-only transaction:", err)
	_ = tx.Commit()

	var name string
	_ = db.QueryRowContext(ctx, "SELECT name FROM users WHERE id=1").Scan(&name)
	fmt.Println("name is now:", name)

	for _, lv := range []sql.IsolationLevel{sql.LevelReadUncommitted, sql.LevelRepeatableRead, sql.LevelSnapshot, sql.LevelLinearizable} {
		t, e := db.BeginTx(ctx, &sql.TxOptions{Isolation: lv})
		fmt.Printf("begin with %-18s -> %v\n", lv, e)
		if e == nil {
			_ = t.Rollback()
		}
	}
}
```

```
begin read-only: <nil>
write inside the read-only transaction: <nil>
name is now: written
begin with Read Uncommitted   -> <nil>
begin with Repeatable Read    -> <nil>
begin with Snapshot           -> <nil>
begin with Linearizable       -> <nil>
```

The difference is that the write inside the read-only transaction should have failed, and that every isolation level other than the default and serializable should have been refused at `BeginTx`.

## Expected behavior

A read-only transaction should refuse a write, and an isolation level SQLite does not provide should be refused when the transaction begins rather than silently downgraded. This is the behavior `database/sql` guarantees on its own: `ctxDriverBegin` returns "sql: driver does not support read-only transactions" and "sql: driver does not support non-default isolation level" for a driver that only implements `driver.Conn.Begin`, and that safety net disappears the moment a driver implements `driver.ConnBeginTx`. Refusing what cannot be honored is also what this package already does elsewhere -- a source it can read but cannot write back is refused by Build rather than written as something else -- so accepting an option and dropping it is the odd one out. `ReadOnly` can be honored rather than refused: SQLite has a per-connection `query_only` pragma, which is the same mechanism `DBBuilder.OpenReadOnly` already uses for a whole handle, so the transaction can set it on entry and clear it when the transaction ends. `LevelDefault` and `LevelSerializable` are what SQLite gives and should keep working unchanged.

## Cause (guess)

The guess is that nothing between the caller and the driver looks at the options. `modernc.org/sqlite`'s `newTx` in tx.go reads `opts.ReadOnly` only to decide whether to prepend its `beginMode` to the BEGIN statement, and never reads `opts.Isolation` at all, so both options reach SQLite as a plain `begin`. This package wraps connections already for auto-save (`autoSaveConnection.BeginTx` in save.go), so that wrapper is where the check belongs; the plain handle and the read-only handle are opened with `sql.Open` and have no wrapper at all today, and the dialect handle's `dialectConnector` does not intercept `BeginTx`. Fixing it in this package rather than upstream keeps the four handle kinds consistent and does not depend on a dependency release.

## Tests to add

- In `concurrency_test.go`, a subtest per handle kind (plain, auto-save, dialect, read-only) that begins a transaction with `ReadOnly: true`, asserts a write inside it fails, asserts a SELECT inside it still succeeds, and asserts the file on disk is unchanged after Close.
- In `concurrency_test.go`, a table over every `sql.IsolationLevel`, asserting that `LevelDefault` and `LevelSerializable` begin and that every other level is refused with an error naming the level. Driving it from the exported list rather than from a hand-written subset is what keeps a level added to `database/sql` later from slipping through unchecked.
- A case that begins a read-only transaction, commits it, and then writes outside a transaction, asserting the write succeeds: the pragma has to be cleared when the transaction ends, or a pooled connection stays read-only for whoever gets it next.

### LINKED ISSUE #850: Auto-save reports success and writes the stale file when the transaction was begun by a BEGIN statement
Found on main (544c281) with Go 1.26.4 on Linux, while walking the auto-save and transaction state transitions of a database this package opens.

## Phenomenon

`EnableAutoSave` refuses to save when a transaction begun with `db.BeginTx` is still open at Close, and says so; a transaction begun by running BEGIN through `db.Exec` is invisible to that guard, so Close reports success and writes the file the loaded rows came from, not the rows the caller last saw. The caller's statements ran, a SELECT on the same handle read them back, and Close returned nil -- and the file on disk still holds what it held before the session. Nothing in the sequence reports a problem, which makes this worse than the case the guard already covers: a caller who reads the returned error learns nothing, because there is no error.

## Reproduction

```go
package main

import (
	"context"
	"fmt"
	"os"
	"path/filepath"

	"github.com/nao1215/filesql"
)

func main() {
	ctx := context.Background()
	dir, _ := os.MkdirTemp("", "beg")
	path := filepath.Join(dir, "users.csv")
	_ = os.WriteFile(path, []byte("id,name\n1,alice\n"), 0o600)

	db, err := filesql.NewBuilder().AddPath(dir).EnableAutoSave("").Open(ctx)
	if err != nil {
		panic(err)
	}
	_, err = db.ExecContext(ctx, "BEGIN")
	fmt.Println("BEGIN:", err)
	_, err = db.ExecContext(ctx, "UPDATE users SET name='written' WHERE id=1")
	fmt.Println("UPDATE:", err)
	var name string
	_ = db.QueryRowContext(ctx, "SELECT name FROM users WHERE id=1").Scan(&name)
	fmt.Println("SELECT sees:", name)
	fmt.Println("Close:", db.Close())
	b, _ := os.ReadFile(path)
	fmt.Printf("users.csv: %q\n", string(b))
}
```

```
BEGIN: <nil>
UPDATE: <nil>
SELECT sees: written
Close: <nil>
users.csv: "id,name\n1,alice\n"
```

The same program with `db.BeginTx` in place of the BEGIN statement prints "filesql: database operation failed: auto-save was skipped because the database was closed while a transaction was still open; commit or roll back before closing" from Close, which is the behavior this one should have.

## Expected behavior

Close should report the skipped save here for the same reason it reports it for `db.BeginTx`: the database is closing with work the caller has neither committed nor rolled back, so what is in it is not a state they asked to keep. Whether the transaction was begun through `database/sql` or by a BEGIN statement does not change that, and the caller cannot tell the two apart from the outcome today. A comment on `autoSaveConnector.transactionBegan` in save.go already names this hole -- "one started by running BEGIN through Exec is invisible here, and a close during one of those still waits forever" -- but the second half of that sentence is not what happens: the close does not wait, it returns nil and writes the stale file, because closing the pooled connection rolls the transaction back before the anchor connection reads the tables.

## Cause (guess)

The guess is that the count `autoSaveConnector.Close` reads is kept only by `autoSaveConnection.BeginTx` and `autoSaveTransaction.Commit`/`Rollback` in save.go, and a BEGIN that arrives as statement text goes through `autoSaveConnection.ExecContext`, which forwards it to the driver without looking at it. The fix is to have that forwarding look: a statement whose first keyword is BEGIN counts the connection's transaction as open, and one whose first keyword is COMMIT, END or ROLLBACK counts it as closed, so the existing guard covers both spellings. Counting conservatively is the right bias -- refusing a save that could have run is recoverable with `DumpDatabase`, while writing rows the caller discarded is not -- so an unbalanced BEGIN should leave the count above zero rather than be guessed back down.

## Tests to add

- In `autosave_test.go`, the reproduction above as a regression test: `EnableAutoSave("")`, BEGIN through Exec, an UPDATE, then Close, asserting Close returns an error that is `ErrDatabaseOperation` and that the file is byte-identical to what was written before the open.
- The balanced sibling in the same file: BEGIN, an UPDATE, COMMIT, then Close, asserting Close returns nil and the file holds the updated row. Without it the fix can be written as "any BEGIN blocks every later save", which passes the first test and breaks the feature.
- The spellings SQLite accepts, as a table over "BEGIN", "BEGIN TRANSACTION", "BEGIN IMMEDIATE", "BEGIN DEFERRED", "BEGIN EXCLUSIVE", with leading whitespace and mixed case, asserting each is counted; and over "COMMIT", "END", "END TRANSACTION", "ROLLBACK", asserting each releases the count.
- A case asserting a statement that merely starts with those letters is not mistaken for one, such as `SELECT 1 AS beginning` and an UPDATE of a column named `commit`, so the check does not fire on a prefix match.

nao1215:
Three fixes found by walking the transaction state transitions of a database this package opens, and by finishing the MySQL comparison work #810 named.

## A sql.TxOptions this package cannot honor is refused rather than dropped

`db.BeginTx(ctx, &sql.TxOptions{ReadOnly: true})` returned a transaction that went on to accept an INSERT, so a caller who declared a transaction read-only as a guard rail had none, and with auto-save on the rows it wrote reached the file at close. Every isolation level was accepted as well, up to `sql.LevelLinearizable`, though SQLite runs serializable transactions and has nothing else to give. `database/sql` refuses both on its own for a driver that implements only the legacy `Begin`, and that safety net disappears the moment a driver implements `ConnBeginTx`. A read-only transaction is now held under SQLite's `query_only` pragma for as long as it runs, and the permission comes back when it ends -- including on a handle from `OpenReadOnly`, which was read-only already and stays so. A level other than `sql.LevelDefault` and `sql.LevelSerializable` is refused with `ErrDatabaseOperation` naming it.

## Auto-save no longer reports success and writes the stale file after a BEGIN statement

A transaction begun with `db.BeginTx` and left open at close already stopped the save and said so; one begun by running BEGIN through `Exec` was invisible to that guard, so Close returned nil and wrote the rows the file already held. The caller's statements had run and a SELECT on the same handle read them back, and closing the pooled connection rolled them back before the save read the tables. A BEGIN run as a statement now counts the connection's transaction as open, and COMMIT, END and a bare ROLLBACK release it; `ROLLBACK TO` a savepoint does not, since it leaves the transaction around it open. The count is deliberately biased toward refusing: a save that was skipped is recoverable with `DumpDatabase`, and rows written over the caller's own are not.

## MySQL's NULLIF compares the way MySQL compares

SQLite's own `nullif()` was answering, and it compares by storage class, so `NULLIF('abc', 0)` answered 'abc' where MySQL answers NULL and `NULLIF('1', 1)` answered '1'. The comparison now follows MySQL's "=", which is the reading FIELD here is already on. Checked against mysql:8.4 over every pair of a thirty-value matrix; the only remaining differences are the spelling of a decimal literal written in the query, which SQLite has no exact decimal to hold.

## Shape of the change

The first two are one wrapper. All four handle kinds -- plain, auto-save, dialect and read-only -- now open their connections through `guardedConn` in the new `transaction.go`, where before two of them were bare `sql.Open` handles and the auto-save one had a wrapper of its own. Folding the auto-save connection and transaction into it also puts back the optional interfaces that wrapper dropped: `PrepareContext`, `Ping`, `ResetSession` and `IsValid`.

## Verification

`make test`, `make lint`, `make lint-staticcheck` and `go test -race ./...` are clean, and `GOOS=windows` and `GOOS=darwin` build. sqly builds and its tests pass against this checkout through a local replace. Benchmarks over three runs each show no regression: geomean -1.28% on time and -0.10% on bytes per operation.

Closes #849
Closes #850
Refs #810

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction handling, including read-only and supported isolation-level enforcement.
  - Auto-save now detects transactions started by SQL statements or prepared statements and avoids saving while transactions remain open.
  - Corrected savepoint rollback handling so open transactions are tracked accurately.
  - Prevented failed saves from being reported as successful.
  - Corrected MySQL `NULLIF` behavior to match MySQL comparison rules, including numeric, text, and `NULL` values.

- **Documentation**
  - Added guidance on SQLite transaction behavior, isolation levels, read-only transactions, and auto-save limitations.


nao1215:
Both fixed in 8caaccc.

A prepared statement whose keyword opens or closes a transaction is now wrapped so it carries the reading; every other statement is left as the driver's own, so the wrapper is not in the way of a query. Regression tests in `autosave_test.go`: a prepared BEGIN left open makes Close refuse the save and leaves the file untouched, and a prepared BEGIN, INSERT and COMMIT still saves.

`ROLLBACK TRANSACTION TO` now skips the optional keyword before looking for TO, with the spelling added to `TestStatementTxEffect` and an end-to-end case in `autosave_test.go` that rolls back to a savepoint and asserts the save is still refused.

PR #888, opened 2026-08-31
fix: refuse to read a path that is not a regular file

### LINKED ISSUE #887: ExcelSheetsInFile and CreateReaderForFile block forever on a named pipe
Found on main (11fca36) after the load routes stopped opening a named pipe: the same question asked of the exported calls that open a path without going through a load, and of the read an in-place save does before it writes.

## Phenomenon

`ExcelSheetsInFile` and `CompressionFactory.CreateReaderForFile` open the path they are given without asking what kind of file it is, so both block forever on a named pipe, past any context, the way a load did. `ExcelSheetsInFile` is the call a caller makes to look at a workbook before deciding to load it, which is exactly the moment they do not yet trust the path. The same open is what an in-place save reads the source through, so a source replaced by a pipe between the load and the save hangs `Close`, which takes no context at all.

## Reproduction

```go
package repro

import (
	"os"
	"os/exec"
	"path/filepath"
	"testing"
	"time"

	"github.com/nao1215/filesql"
)

func TestExportedCallsOnANamedPipe(t *testing.T) {
	dir := t.TempDir()
	book := filepath.Join(dir, "book.xlsx")
	data := filepath.Join(dir, "data.csv")
	for _, p := range []string{book, data} {
		if err := exec.Command("mkfifo", p).Run(); err != nil {
			t.Skipf("this platform has no named pipes: %v", err)
		}
	}

	try := func(label string, run func() error) {
		done := make(chan error, 1)
		go func() { done <- run() }()
		select {
		case err := <-done:
			t.Logf("%-22s %v", label, err)
		case <-time.After(6 * time.Second):
			t.Errorf("%-22s did not return", label)
		}
	}

	try("ExcelSheetsInFile", func() error {
		_, err := filesql.ExcelSheetsInFile(book)
		return err
	})
	try("CreateReaderForFile", func() error {
		_, cleanup, err := filesql.NewCompressionFactory().CreateReaderForFile(data)
		if cleanup != nil {
			_ = cleanup()
		}
		return err
	})
	_ = os.Remove
}
```

```
--- FAIL: TestExportedCallsOnANamedPipe (12.00s)
    main_test.go:36: ExcelSheetsInFile      did not return
    main_test.go:36: CreateReaderForFile    did not return
```

`ExcelSheetTableNames` returns promptly, which is right: it works out names and never opens the file.

The save-time shape, with nothing exotic in the API: load a CSV with `EnableAutoSave("")`, change a row, replace the source with a named pipe, and `Close` never returns. A save reads the source before it writes -- for its compression, its text encoding and its line terminator -- and that read is this same open.

## Expected behavior

A path this package opens for reading should be refused when it is not a regular file, named and with its kind, which is the answer a load now gives. `ErrUnsupportedFormat` is the sentinel the load side uses and the wording is `atomicwrite.DescribeFileMode`, so all of them read alike. A caller who really wants to read a pipe can open it themselves and hand the reader to `CompressionHandler.CreateReader` or to `DBBuilder.AddReader`, where blocking is their own choice and the package documentation already says a read blocked inside a source cannot be interrupted.

## Cause (guess)

This is a guess from reading the code. `CompressionFactory.CreateReaderForFile` in `compression.go` calls `os.Open(path)` with nothing between, and it is the one place every read of a path goes through: `save.go` reads the source there before an in-place write, `line_ending.go` and `encoding.go` read it there to keep the terminator and the encoding, `excel_sheet.go` reads a workbook there for `ExcelSheetsInFile`, and `stream_processor.go` reads a compressed source there during a load. So one check inside it covers every one of them, including the two exported calls, rather than being added at five call sites. The load-side checks added earlier sit further out, in the collection, which is where a refusal can name the entry before anything is staged; they stay, and this is the floor under them.

## Tests to add

- In `compression_test.go`, `CreateReaderForFile` on a named pipe is refused rather than blocking, with the call on a goroutine under a wall-clock limit so a block fails the test instead of taking the package timeout.
- In `excel_sheet_test.go`, `ExcelSheetsInFile` on a named pipe is refused, since that is the exported call a caller reaches first.
- In `autosave_test.go`, a source replaced by a named pipe after the load makes `Close` report rather than hang, which is the save-time read.
- The siblings that must keep working: a regular file, a compressed regular file, and a regular file reached through a symbolic link all still open, so the check follows the link rather than refusing it.
- A case that pins the division: `DBBuilder.AddReader` with a reader of the caller's own is untouched, because a reader the caller opened is theirs to block on.

nao1215:
The load routes stopped opening a named pipe. This is the same question asked of the exported calls that reach a path without going through a load, and of the read an in-place save does before it writes.

`ExcelSheetsInFile` and `CompressionFactory.CreateReaderForFile` open the path they are given without asking what kind of file it is, so both block forever on a named pipe, past any context. `ExcelSheetsInFile` is the call a caller makes to look at a workbook before deciding to load it — exactly the moment they do not yet trust the path. And the same open is what an in-place save reads the source through, for its compression, its text encoding and its line terminator, so a source replaced by a pipe between the load and the save hangs `Close`, which takes no context at all.

`CreateReaderForFile` is the one place every read of a path in this package goes through — a load of a compressed source, the save's read of the original, `ExcelSheetsInFile` — so the check sits there rather than at five call sites. A load still refuses such a path earlier, while collecting, where it can name the entry before anything is opened; this is the floor under that. A caller who means to read a pipe opens it themselves and hands the reader to `CompressionHandler.CreateReader` or `DBBuilder.AddReader`, where the blocking is their own choice and the package documentation already says a read blocked inside a source cannot be interrupted.

Each of the three ways a caller arrives is covered separately, since each is a different caller: the choke point, the exported call, and the save. Each runs on a goroutine under a wall-clock limit, because a block there has no deadline of its own and a test that waits inline takes the package timeout instead of failing. The save test deliberately does not pin which layer refuses — the read no longer blocks, so the save reaches the write, which refuses a destination that is not a file; what it pins is that `Close` returns and says what is wrong.

Cost: one stat before an open. Measured on the package's own load benchmarks at `-count=5`, plain and for `.gz`, `.zst` and `.xz`, every difference is inside the run-to-run noise and the allocation counts are unchanged.

Verification: `make test` and `make lint` are green, the package cross-builds for windows/amd64, and sqly builds and its tests pass against this checkout through a local replace.

Closes #887


nao1215:
Extended after opening this: the compression factory was not the only read that reached a path directly.

A Parquet source replaced by a named pipe still hung `Close` after the factory was fixed, because `dump_parquet.go` opens the source itself to take its schema before an in-place save replaces it, and never goes through the factory. The same is true of the two reads that rebuild an ACH or a Fedwire structure from the recorded source path in `source_registry.go`.

All of them go through one `openRegularFile` now, which stats before it opens, so a kind is refused once wherever a path is read rather than at each call site. The regression test is a table over the formats an in-place save can write, since each reads its source through a different call: csv through the factory, parquet through its own open, ach through the recorded path. Removing any of the three fixes makes it hang again, which is how I checked that each case reaches what it is meant to.

The Parquet seed is dumped inside the test rather than added to testdata.

PR #892, opened 2026-08-31
fix(prep): compare integers as integers, and read excluded_unless the way required_unless reads it

### LINKED ISSUE #889: excluded_unless naming two pairs forbids a value unless both match, while required_unless asks for one unless either does
Found on main at 5ccd8d9 by reading the conditional-required family of `prep` against the validator dialect it names, and against its own mirror.

## Phenomenon

A `excluded_unless` tag that names more than one field and value pair forbids a value unless every pair matches, while `required_unless` -- the tag `doc.go` says it mirrors -- asks for a value unless any one pair matches. The two are supposed to be the same condition read the other way, so on the same row of the same file they should agree about whether the condition holds; they do not. The rows that disagree are the ones where some pairs match and others do not, which is every row of a real file except the two extremes, so a caller who writes `excluded_unless=Kind paid Tier gold` to mean "a note is only allowed on a paid or gold row" gets every gold row of a free account rejected, and no error tells them the tag was read the other way.

## Reproduction

```go
package main

import (
	"fmt"
	"strings"

	"github.com/nao1215/filesql/prep"
)

type excluded struct {
	Kind string `name:"kind"`
	Tier string `name:"tier"`
	Note string `name:"note" validate:"excluded_unless=Kind paid Tier gold"`
}

type required struct {
	Kind string `name:"kind"`
	Tier string `name:"tier"`
	Note string `name:"note" validate:"required_unless=Kind paid Tier gold"`
}

func main() {
	csv := "kind,tier,note\npaid,gold,x\npaid,silver,x\nfree,gold,x\nfree,silver,x\n"
	var excludedRows []excluded
	_, excludedResult, _ := prep.NewProcessor(prep.FileTypeCSV).Process(strings.NewReader(csv), &excludedRows)
	fmt.Println("excluded_unless rejects:")
	for _, e := range excludedResult.Errors {
		fmt.Println(" ", e)
	}

	blank := "kind,tier,note\npaid,gold,\npaid,silver,\nfree,gold,\nfree,silver,\n"
	var requiredRows []required
	_, requiredResult, _ := prep.NewProcessor(prep.FileTypeCSV).Process(strings.NewReader(blank), &requiredRows)
	fmt.Println("required_unless rejects:")
	for _, e := range requiredResult.Errors {
		fmt.Println(" ", e)
	}
}
```

```
excluded_unless rejects:
  row 2, column "note" (field Note): value must be empty unless Kind is paid and Tier is gold (value="x", tag=excluded_unless)
  row 3, column "note" (field Note): value must be empty unless Kind is paid and Tier is gold (value="x", tag=excluded_unless)
  row 4, column "note" (field Note): value must be empty unless Kind is paid and Tier is gold (value="x", tag=excluded_unless)
```

`required_unless` treats rows 2 and 3 as rows where the condition holds and asks for nothing; `excluded_unless` treats the same two rows as rows where the condition fails and forbids a value. Only row 4, where neither pair matches, should be rejected by `excluded_unless`.

## Expected behavior

`excluded_unless` should forbid a value only when no pair matches, so that it is the exact negation of `required_unless` and reads its parameter the way that tag does. That is what `doc.go` promises of the whole family: "excluded_if, excluded_unless, excluded_with, excluded_with_all, excluded_without and excluded_without_all negate that family: each names the condition under which the cell must be empty, and reads its parameter the way the required tag it mirrors does." It is also what the dialect this package names does -- `excludedUnless` in go-playground/validator returns as soon as one pair matches, exactly like its `requiredUnless`. The sentence in `doc.go` that spells the current behavior, "excluded_unless=Kind paid forbids one unless they all do", is the one that has to change with the code; with a single pair, which is what that example shows, all and any are the same condition, which is why the divergence went unnoticed.

## Cause (guess)

`excludedUnlessValidator.Validate` in `prep/cross_field.go` calls `conditionsMet(v.conditions, targetValues, true)`, asking whether every pair matches, while `requiredUnlessValidator.Validate` next to it calls the same helper with `false`, asking whether any one does. The two calls were written to the same helper but with opposite arguments, and no test names more than one pair for `excluded_unless`, so the pair of tags was never compared on a row where all and any differ.

## Tests to add

- In `prep/cross_field_test.go`, extend the `excluded_unless` table with a two-pair tag over the four rows of the reproduction, asserting that only the row matching neither pair is rejected.
- In `prep/cross_field_test.go`, add a case that runs `required_unless` and `excluded_unless` with the same two-pair parameter over the same rows and asserts that a row is rejected by exactly one of them -- the invariant that the two tags are negations, which catches the sibling mistake in either tag.
- In `prep/cross_field_test.go`, add the same two-pair shape for `excluded_if` against `required_if` so the other conditional pair is held to the same invariant; those two agree today and should keep agreeing.

### LINKED ISSUE #890: A comparison tag on an integer field decides two integers past 2^53 by the doubles they round to
Found on main at 5ccd8d9 by pushing the comparison tags of `prep` past the range a float64 can count in.

## Phenomenon

A comparison tag on an integer field reads both the cell and the tag's parameter as `float64`, so two integers past 2^53 that round to the same double are compared as if they were the same number. `eq=9007199254740993` accepts a cell holding 9007199254740992, `min=9007199254740993` accepts the same cell, `len=9007199254740993` calls it the number it is not, and `gt=9007199254740992` rejects a cell holding 9007199254740993 -- a value that is greater than the threshold. The field is `int64` and every one of these values fits in it exactly, so the caller wrote a bound their data can express and got the answer for a neighbouring number. Nothing reports the loss: a row that should have been rejected is written to the output the processor returns, and a row that should have passed is dropped from it under `WithValidRowsOnly`. Identifiers, transaction amounts in minor units and epoch nanoseconds all live past 2^53, so this is reachable with ordinary data.

## Reproduction

```go
package main

import (
	"fmt"
	"strings"

	"github.com/nao1215/filesql/prep"
)

type row struct {
	Eq  int64 `name:"eq" validate:"eq=9007199254740993"`
	Gt  int64 `name:"gt" validate:"gt=9007199254740992"`
	Min int64 `name:"min" validate:"min=9007199254740993"`
	Len int64 `name:"len" validate:"len=9007199254740993"`
}

func main() {
	csv := "eq,gt,min,len\n9007199254740992,9007199254740993,9007199254740992,9007199254740992\n"
	var rows []row
	_, result, err := prep.NewProcessor(prep.FileTypeCSV).Process(strings.NewReader(csv), &rows)
	fmt.Println("err:", err)
	fmt.Println("valid:", result.ValidRowCount, "invalid:", result.InvalidRowCount())
	for _, e := range result.Errors {
		fmt.Println(" ", e)
	}
}
```

```
err: <nil>
valid: 0 invalid: 1
  row 1, column "gt" (field Gt): value must be greater than 9007199254740992 (value="9007199254740993", tag=gt)
```

Three of the four tags answered the wrong way silently: `eq`, `min` and `len` accepted 9007199254740992 against a parameter of 9007199254740993, and the one error that was reported is the one comparison that should have passed. Note also that the message spells its own threshold as 9007199254740992, because the parameter was rounded on the way in.

## Expected behavior

A comparison between two integers should be decided by the integers. `eq=9007199254740993` should reject 9007199254740992, `min=9007199254740993` and `len=9007199254740993` should reject it too, and `gt=9007199254740992` should accept 9007199254740993, and the message a rejection carries should spell the parameter the caller wrote rather than the double it rounded to. This is what `doc.go` describes -- "on any other field all of them compare the numeric value and len means the value equals the parameter" -- and what go-playground/validator does, since it compares an `int64` field with `field.Int()` against a parameter parsed with `asInt`. A cell or a parameter that spells a fraction or an exponent still has to go through `float64`; only the case where both sides spell an integer changes.

## Cause (guess)

Every numeric comparison in `prep/validate.go` stores its parameter as a `threshold float64` and reads the cell with `strconv.ParseFloat`: `equalValidator`, `notEqualValidator`, `greaterThanValidator`, `greaterThanEqualValidator`, `lessThanValidator`, `lessThanEqualValidator`, `minValidator` and `maxValidator` all follow that shape, and `lengthValidator` compares `f != float64(v.length)` on a numeric field. The parameter is parsed once in `parser.go` through `buildFloatValidator`, and in `specializeValidator` for the deferred `eq`/`ne` pair, so the rounding happens before the cell is ever read. The fix belongs in one place: a single helper that orders two numbers written as text, comparing them as `int64` when both spell one and falling back to `float64` otherwise, used by all nine validators.

## Tests to add

- In `prep/validate_test.go`, add the four tags of the reproduction against the 2^53 boundary: `eq`, `min` and `len` must reject 9007199254740992 against a parameter of 9007199254740993, and `gt=9007199254740992` must accept 9007199254740993.
- In `prep/validate_test.go`, add the same boundary for the sibling tags the same helper serves -- `ne`, `gte`, `lt`, `lte` and `max` -- so a fix that reaches only the tags named above is caught.
- In `prep/validate_test.go`, add the negative side of the boundary: `math.MinInt64` and `math.MinInt64 + 1` compared against each other, which round to the same double the same way.
- In `prep/validate_test.go`, keep a case where one side spells a fraction (`gt=1.5` against `1.4` and `1.6`) and one where a parameter is written in exponent form (`eq=1e3` against `1000`), so the float fallback is not lost with the fix.
- In `prep/validate_test.go`, assert the message a rejected large parameter carries spells the parameter as written rather than the rounded double.

### LINKED ISSUE #891: A cross-field comparison decides two integers past 2^53 by the doubles they round to
Found on main at 5ccd8d9 alongside the single-field comparison tags, by pushing the cross-field comparisons past the range a float64 can count in.

## Phenomenon

A cross-field comparison on a non-string field reads both cells as `float64`, so two integers past 2^53 that round to the same double are ordered as if they were the same number. With an `int64` field, `nefield` reports that 9007199254740993 equals 9007199254740992 and `gtfield` reports that 9007199254740993 is not greater than it. Both cells came out of the file exactly and both fit the field they landed in, so the caller gets a verdict about a neighbouring pair of numbers with nothing to say it happened: a row that should have passed is reported as an error and, under `WithValidRowsOnly`, dropped from the reader the processor returns. Two identifiers that differ, or a pair of amounts in minor units, are enough to reach it.

## Reproduction

```go
package main

import (
	"fmt"
	"strings"

	"github.com/nao1215/filesql/prep"
)

type row struct {
	A  int64 `name:"a"`
	Ne int64 `name:"ne" validate:"nefield=A"`
	Gt int64 `name:"gt" validate:"gtfield=A"`
}

func main() {
	csv := "a,ne,gt\n9007199254740992,9007199254740993,9007199254740993\n"
	var rows []row
	_, result, err := prep.NewProcessor(prep.FileTypeCSV).Process(strings.NewReader(csv), &rows)
	fmt.Println("err:", err)
	fmt.Println("valid:", result.ValidRowCount, "invalid:", result.InvalidRowCount())
	for _, e := range result.Errors {
		fmt.Println(" ", e)
	}
}
```

```
err: <nil>
valid: 0 invalid: 1
  row 1, column "ne" (field Ne): value must not equal field A (value="9007199254740993", tag=nefield)
  row 1, column "gt" (field Gt): value must be greater than field A (value="9007199254740993", tag=gtfield)
```

9007199254740993 does not equal 9007199254740992 and is greater than it, so neither error should have been reported and the row should have counted as valid.

## Expected behavior

A comparison between two cells that both spell integers should be decided by the integers, so `nefield` passes and `gtfield` passes on the row above, and `eqfield` on the same pair reports them as different. Cells that spell a fraction or an exponent, or that spell no number at all, keep the behavior they have now -- the double, then the text. `doc.go` says the cross-field comparison tags "follow the field the same way" as the single-field ones, which on a non-numeric field means comparing the numeric value; go-playground/validator compares two `int64` fields with `field.Int()` on both sides and never goes through a double.

## Cause (guess)

`baseCrossFieldValidator.compare` in `prep/cross_field.go` reads both cells with `strconv.ParseFloat` whenever `comparesText` is false, which is every field kind but string. The rounding is in that one function, so the fix is in that one function: order the two as `int64` when both spell one, keep the `float64` comparison when either does not, and keep the text fallback for a pair that spells no number at all. It is the same defect as the one in the single-field comparison tags, in the second of the two places this package compares numbers, and the fix should leave one helper that both call rather than two that agree by inspection.

## Tests to add

- In `prep/cross_field_test.go`, add the reproduction: an `int64` field pair holding 9007199254740993 and 9007199254740992, asserting `nefield` and `gtfield` pass and `eqfield` fails.
- In `prep/cross_field_test.go`, add the same boundary for the remaining comparison tags -- `gtefield`, `ltfield` and `ltefield` -- so a fix that reaches only the two named above is caught.
- In `prep/cross_field_test.go`, add the negative side: `math.MinInt64` against `math.MinInt64 + 1`.
- In `prep/cross_field_test.go`, keep the existing fallbacks covered: a pair of cells spelling fractions, and a pair on a string field where the tag orders the strings rather than the numbers.
- Once both this and the single-field comparison read numbers through one helper, add a case that runs the same pair of values through `gt=` and through `gtfield=` and asserts the two answers agree, which is the invariant that keeps the two places from drifting again.

nao1215:
Three defects in `prep`, found by reading the comparison tags and the conditional-required family against the validator dialect the package names and against each other.

## What was wrong

A comparison tag on an integer field held its parameter as a `float64` and read the cell with `strconv.ParseFloat`, so two `int64` values past 2^53 that round to the same double were compared as one number: `eq=9007199254740993` accepted a cell holding 9007199254740992, `min` and `len` accepted it too, and `gt=9007199254740992` refused 9007199254740993 -- a value that is greater than the threshold ([#890](https://github.com/nao1215/filesql/issues/890)). The cross-field comparisons did the same on their two cells, so `nefield` called two different integers equal and `gtfield` called the greater one not greater ([#891](https://github.com/nao1215/filesql/issues/891)). Neither reported the rounding: a row that should have been refused went into the reader the processor returns, and a row that should have passed was dropped from it under `WithValidRowsOnly`.

`excluded_unless` naming more than one field and value pair forbade a value unless every pair matched, while `required_unless` -- the tag `doc.go` says it negates -- asks for a value unless any one pair matches ([#889](https://github.com/nao1215/filesql/issues/889)). The two disagreed on every row where some pairs match and others do not, so `excluded_unless=Kind paid Tier gold`, written to mean that a note is allowed on a paid or a gold row, refused the note on every gold row of a free account. `excludedUnless` in go-playground/validator returns as soon as one pair matches, exactly like its `requiredUnless`.

## What changed

A number a cell or a tag parameter spells is now read through one type that keeps both the double and the integer, and a comparison between two integers is decided by the integers whatever their size. Both places this package compares numbers go through it, so the comparison tags and the cross-field comparisons cannot drift apart again, and a rejection quotes the parameter as the caller wrote it rather than the double it rounded to. A cell or a parameter that spells a fraction or an exponent still goes through `float64`, a comparison on a string field still orders the strings, and a pair that spells no number at all still falls back to the text.

`excluded_unless` now allows a value as soon as one pair matches, and the message a refusal carries joins its conditions with "or" rather than "and". A tag naming no pair still asks for nothing. This changes which rows the tag refuses for any tag naming two or more pairs; a tag naming one pair is unaffected, since all and any are the same condition there. `doc.go` is updated in both places it described the old rule.

## How it is kept fixed

`prep/validate_test.go` runs all nine comparison tags at the 2^53 boundary and at the negative end, keeps the fractional and exponent parameters covered so the float fallback is not lost, and holds a message to the parameter the caller wrote. `prep/cross_field_test.go` does the same for all six cross-field comparisons, holds `gt=` and `gtfield=` to the same answer on the same pair of values, and holds `excluded_unless` and `required_unless` -- and `excluded_if` and `required_if`, the pair that already agreed -- to refusing a row from exactly one side each.

No public symbol was added, removed or changed. `make test`, `make lint` and `make lint-staticcheck` are clean, and sqly builds and tests against this branch.

The shape of a number is checked before `strconv.ParseInt` is called, because a failed parse returns an error that escapes to the heap and a validation chain runs on every cell of every row. With that check the change is not slower than what it replaces: over `-count=5`, `BenchmarkValidatorsOnly` holds at 264 B/op and 10 allocs/op with ns/op unchanged, and `BenchmarkNumericValidation`, `BenchmarkComplexValidatorChain` and `BenchmarkProcessCrossFieldCSV` each come out faster, since an integer cell no longer goes through `ParseFloat`.

Closes #889
Closes #890
Closes #891

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed numeric comparisons so large integers are compared exactly, including equality, ordering, range, and length validations.
  * Corrected `excluded_unless` to allow values when any specified condition matches.
  * Preserved the original numeric formatting in validation error messages, including leading zeros, signs, and exponents.

* **Documentation**
  * Clarified integer comparison behavior beyond floating-point precision.
  * Clarified how multi-condition validation tags evaluate matching conditions.


nao1215:
Both taken, and the misspellings with them.

The threshold now carries the text it was read from and `String` returns it, so `eq=1e3` answers with `1e3`, `min=0010` with `0010`, and `eq=+001` with `+001`. `prep/validate_test.go` pins three of those spellings.

The documentation no longer says "range": it now says a pair past the point where a float64 stops spelling every integer exactly is ordered rather than rounded onto one number.

PR #920, opened 2026-09-01
fix: read the words a statement is written with, and refuse an export that names nothing

### LINKED ISSUE #916: A PostgreSQL ONLY before a table name is read as the table itself, so a query, an UPDATE and a DELETE name the wrong table
Found on main (c5fb518) while sweeping the spellings a dialect accepts for a construct this package already refuses or rewrites by name.

## Phenomenon

In the PostgreSQL dialect, the keyword ONLY in front of a table name is read as the table name itself and the real table name is read as its alias, so `dialect.Translate` turns `SELECT * FROM ONLY t` into `SELECT * FROM ONLY AS t`, `UPDATE ONLY t SET a = 1` into `UPDATE ONLY AS t SET a = 1` and `DELETE FROM ONLY t` into `DELETE FROM ONLY AS t`. ONLY in PostgreSQL means "this table and not the tables that inherit from it", and since a filesql database has no inheritance it names exactly the table that follows it. A caller who writes it gets `no such table: ONLY` on a database without such a table, and on a database that has one -- a file named `only.csv` loads as the table `only`, and SQLite matches a table name without regard to case -- the query silently answers from, updates or deletes the wrong table. The parenthesized spelling is worse still: `SELECT * FROM ONLY (t)` renders as `SELECT * FROM ONLY(t)`, a call to a table-valued function.

## Reproduction

```go
package main

import (
	"fmt"

)

func main() {
	for _, q := range []string{
		"SELECT * FROM ONLY t",
		"SELECT * FROM ONLY t AS x",
		"SELECT * FROM ONLY (t)",
		"SELECT * FROM t *",
		"UPDATE ONLY t SET a = 1",
		"DELETE FROM ONLY t",
		"ALTER TABLE ONLY t ADD COLUMN a INT",
	} {
		out, err := dialect.Translate(dialect.PostgreSQL, q)
		fmt.Printf("%-38s -> %-40q %v\n", q, out, err)
	}
}
```

```
SELECT * FROM ONLY t                   -> "SELECT * FROM ONLY AS t"                <nil>
SELECT * FROM ONLY t AS x              -> ""                                       dialect: invalid SQL syntax: expected end of statement, found AS at line 1, column 22
SELECT * FROM ONLY (t)                 -> "SELECT * FROM ONLY(t)"                  <nil>
SELECT * FROM t *                      -> ""                                       dialect: invalid SQL syntax: expected end of statement, found * at line 1, column 17
UPDATE ONLY t SET a = 1                -> "UPDATE ONLY AS t SET a = 1"             <nil>
DELETE FROM ONLY t                     -> "DELETE FROM ONLY AS t"                  <nil>
ALTER TABLE ONLY t ADD COLUMN a INT    -> ""                                       dialect: SQL feature not implemented by this package: this ALTER TABLE is not implemented at line 1, column 18
```

PostgreSQL 17 answers all five of the first spellings from the table `t`, which is what the caller wrote:

```
$ docker run --rm -d --name pg -e POSTGRES_PASSWORD=pw -e POSTGRES_DB=probe postgres:17-alpine
$ docker exec pg psql -U postgres -d probe -c "CREATE TABLE t(a int); INSERT INTO t VALUES (1);" -c "SELECT * FROM ONLY t;" -c "SELECT * FROM ONLY t AS x;" -c "SELECT * FROM ONLY (t);" -c "SELECT * FROM t *;" -c "UPDATE ONLY t SET a = 2 RETURNING *;"
```

The difference is that ONLY is a keyword of the table reference and this package reads it as a name.

## Expected behavior

In the PostgreSQL dialect, ONLY before a table name -- with or without parentheses around the name -- names that table, so `SELECT * FROM ONLY t` translates to `SELECT * FROM t`, and the alias, the join and the rest of the statement attach to `t` as they do without the keyword. The trailing `*` after a table name, which is the opposite spelling and asks for the inheriting tables too, names the same single table here and can be read and dropped the same way. The keyword belongs to the PostgreSQL grammar alone: MySQL and GoogleSQL have no ONLY in a table reference, `FROM ONLY t` there is a table named ONLY with the alias t, and those dialects have to keep reading it that way. `ALTER TABLE ONLY t` is the same keyword in the same position and should alter `t` rather than report an ALTER TABLE this package does not implement.

## Cause (guess)

`parseTableFactor` in dialect/internal/parser/from.go reads a name and then an optional alias, and nothing looks for ONLY first, so the keyword takes the name's place; `parseTableNameOnly`, which dialect/internal/parser/ddl.go calls for ALTER TABLE, has the same hole and then reports the leftover words as an ALTER TABLE it cannot read. The parenthesized form goes down the table-valued function branch because a name followed by `(` is a call.

## Tests to add

- In dialect/dialect_test.go, add a case to a table-driven test that translates the six PostgreSQL spellings (`FROM ONLY t`, `FROM ONLY t AS x`, `FROM ONLY (t)`, `FROM t *`, `UPDATE ONLY t SET a = 1`, `DELETE FROM ONLY t`) and asserts the rendered SQL names the table `t` and no table named ONLY, and that `ALTER TABLE ONLY t ADD COLUMN a INT` renders the same statement `ALTER TABLE t ADD COLUMN a INT` does.
- In the same file, assert the sibling that must not change: `Translate(MySQL, "SELECT * FROM ONLY t")` and the GoogleSQL one keep reading ONLY as a table name with the alias t, since neither dialect has the keyword.
- In dialect/dialect_test.go, extend the test that prepares translated SQL against SQLite so a query written with ONLY over a real table answers the same rows as the same query without it; a translation that names the wrong table cannot pass that.

### LINKED ISSUE #917: ALTER TABLE spellings that are valid in their dialect are refused as unreadable SQL rather than as unsupported
Found on main (c5fb518) while sweeping the spellings a dialect accepts for a construct this package already refuses by name.

## Phenomenon

Six ALTER TABLE spellings that are valid in the dialect they are written in are refused with ErrInvalidSyntax, which this package documents as "the query could not be read: an unterminated string, a stray parenthesis, an expression that stops in the middle". The caller's SQL is none of those: it is well formed and it asks for something SQLite cannot do. The error a caller gets tells them to fix their typing when there is nothing to fix, and it hides the difference this package's three sentinels exist to draw -- a caller who branches on ErrInvalidSyntax to report a syntax error and on ErrUnsupportedSyntax to fall back reports the wrong one. The same constructs spelled slightly differently are already refused correctly by name (`ALTER TABLE ... SET`, `MODIFY`, `CHANGE`, `ADD CONSTRAINT`), so the classification depends on which spelling the caller happened to use.

## Reproduction

```go
package main

import (
	"fmt"

)

func main() {
	cases := []struct {
		d dialect.Dialect
		q string
	}{
		{dialect.PostgreSQL, "ALTER TABLE t ADD COLUMN a INT, ADD COLUMN b INT"},
		{dialect.PostgreSQL, "ALTER TABLE t RENAME COLUMN a TO b, RENAME COLUMN c TO d"},
		{dialect.PostgreSQL, "ALTER TABLE t ADD COLUMN IF NOT EXISTS a INT"},
		{dialect.PostgreSQL, "ALTER TABLE t DROP COLUMN IF EXISTS a"},
		{dialect.PostgreSQL, "ALTER TABLE t DROP COLUMN a CASCADE"},
		{dialect.MySQL, "ALTER TABLE t ADD COLUMN a INT AFTER b"},
		{dialect.MySQL, "ALTER TABLE t ADD COLUMN a INT FIRST"},
		{dialect.PostgreSQL, "ALTER TABLE IF EXISTS t ADD COLUMN a INT"},
	}
	for _, c := range cases {
		_, err := dialect.Translate(c.d, c.q)
		fmt.Printf("%-11v %-56s -> %v\n", c.d, c.q, err)
	}
}
```

```
postgresql  ALTER TABLE t ADD COLUMN a INT, ADD COLUMN b INT         -> dialect: invalid SQL syntax: expected end of statement, found , at line 1, column 31
postgresql  ALTER TABLE t RENAME COLUMN a TO b, RENAME COLUMN c TO d -> dialect: invalid SQL syntax: expected end of statement, found , at line 1, column 35
postgresql  ALTER TABLE t ADD COLUMN IF NOT EXISTS a INT             -> dialect: invalid SQL syntax: expected end of statement, found NOT at line 1, column 29
postgresql  ALTER TABLE t DROP COLUMN IF EXISTS a                    -> dialect: invalid SQL syntax: expected end of statement, found EXISTS at line 1, column 30
postgresql  ALTER TABLE t DROP COLUMN a CASCADE                      -> dialect: invalid SQL syntax: expected end of statement, found CASCADE at line 1, column 29
mysql       ALTER TABLE t ADD COLUMN a INT AFTER b                   -> dialect: invalid SQL syntax: expected end of statement, found AFTER at line 1, column 32
mysql       ALTER TABLE t ADD COLUMN a INT FIRST                     -> dialect: invalid SQL syntax: expected end of statement, found FIRST at line 1, column 32
postgresql  ALTER TABLE IF EXISTS t ADD COLUMN a INT                 -> dialect: SQL feature not implemented by this package: this ALTER TABLE is not implemented at line 1, column 16
```

The difference is that each of these reads as "the caller wrote something unreadable" where the caller wrote a statement their engine accepts.

## Expected behavior

Each spelling is read to its end and then refused by name, with the sentinel that says which kind of refusal it is. A list of actions in one ALTER TABLE and a column placed with FIRST or AFTER are valid statements SQLite cannot express, so they belong to ErrUnsupportedSyntax with a message naming the construct: SQLite makes one change per ALTER TABLE, and it appends a column rather than placing it. IF NOT EXISTS on ADD COLUMN and IF EXISTS on DROP COLUMN are the same: SQLite has neither, and skipping the check would run the statement the caller asked to skip. `ALTER TABLE IF EXISTS t` should say so by name instead of reporting an ALTER TABLE that is not implemented at a column in the middle of the keyword. CASCADE and RESTRICT after a dropped column are the one exception and should be read and dropped, because ALTER TABLE DROP COLUMN already does that in this package's DROP TABLE, which documents that "SQLite has neither word and behaves as RESTRICT does not"; the two statements deserve the same answer.

## Cause (guess)

`parseAlter` in dialect/internal/parser/ddl.go returns as soon as it has read one action and leaves whatever follows to the caller, which reports the leftover token as an unexpected end of statement; the ADD and DROP branches read a column definition or a name with nothing in front of them looking for IF NOT EXISTS, IF EXISTS, FIRST, AFTER, CASCADE or a comma, and the IF EXISTS after ALTER TABLE is read as the table's name so the switch falls through to its default.

## Tests to add

- In dialect/dialect_test.go, extend the test that covers ALTER TABLE refusals with the seven spellings above, asserting for each the sentinel (ErrUnsupportedSyntax) and that the message names the construct rather than a token position, and that none of them is ErrInvalidSyntax.
- In the same test, assert `ALTER TABLE t DROP COLUMN a CASCADE` and `... RESTRICT` translate to the same SQL as `ALTER TABLE t DROP COLUMN a`, which is what DROP TABLE already does with those words.
- Add the sibling that keeps the class from coming back on the other side: a genuinely unreadable ALTER TABLE (`ALTER TABLE t ADD COLUMN a INT,`, ending after the comma) stays ErrInvalidSyntax.

### LINKED ISSUE #918: MySQL LOCK IN SHARE MODE after a bare table name and PostgreSQL FOR KEY SHARE are refused as unreadable SQL
Found on main (c5fb518) while sweeping the spellings a dialect accepts for a construct this package already refuses by name.

## Phenomenon

`dialect.Translate` refuses a row-locking clause with ErrUnsupportedSyntax and the sentence "a row-locking clause is not supported; SQLite locks the database rather than rows", but two spellings never reach it and come back as ErrInvalidSyntax instead. MySQL's `SELECT * FROM t LOCK IN SHARE MODE` is refused as unreadable SQL when the clause follows the table name directly, and the same query with an alias or a WHERE in between is refused correctly, so the classification depends on what stands before the clause. PostgreSQL's `FOR KEY SHARE` is refused as unreadable in every position, while its three siblings -- `FOR UPDATE`, `FOR NO KEY UPDATE`, `FOR SHARE` -- are refused correctly. A caller who branches on ErrInvalidSyntax to report a typing mistake and on ErrUnsupportedSyntax to drop the clause and retry gets the wrong branch for a statement their own engine runs.

## Reproduction

```go
package main

import (
	"fmt"

)

func main() {
	cases := []struct {
		d dialect.Dialect
		q string
	}{
		{dialect.MySQL, "SELECT * FROM t LOCK IN SHARE MODE"},
		{dialect.MySQL, "SELECT * FROM t AS x LOCK IN SHARE MODE"},
		{dialect.MySQL, "SELECT * FROM t WHERE a = 1 LOCK IN SHARE MODE"},
		{dialect.PostgreSQL, "SELECT * FROM t FOR KEY SHARE"},
		{dialect.PostgreSQL, "SELECT * FROM t FOR NO KEY UPDATE"},
	}
	for _, c := range cases {
		_, err := dialect.Translate(c.d, c.q)
		fmt.Printf("%-11v %-48s -> %v\n", c.d, c.q, err)
	}
}
```

```
mysql       SELECT * FROM t LOCK IN SHARE MODE               -> dialect: invalid SQL syntax: expected end of statement, found IN at line 1, column 22
mysql       SELECT * FROM t AS x LOCK IN SHARE MODE          -> dialect: syntax not supported on SQLite backend: a row-locking clause is not supported; SQLite locks the database rather than rows at line 1, column 22
mysql       SELECT * FROM t WHERE a = 1 LOCK IN SHARE MODE   -> dialect: syntax not supported on SQLite backend: a row-locking clause is not supported; SQLite locks the database rather than rows at line 1, column 29
postgresql  SELECT * FROM t FOR KEY SHARE                    -> dialect: invalid SQL syntax: expected end of statement, found FOR at line 1, column 17
postgresql  SELECT * FROM t FOR NO KEY UPDATE                -> dialect: syntax not supported on SQLite backend: a row-locking clause is not supported; SQLite locks the database rather than rows at line 1, column 17
```

MySQL 8.4 runs the first query, so LOCK there is a keyword and not the table's alias:

```
$ docker run --rm -d --name my -e MYSQL_ROOT_PASSWORD=pw -e MYSQL_DATABASE=probe mysql:8.4
$ docker exec my mysql -uroot -ppw -D probe -e "CREATE TABLE t(a int); INSERT INTO t VALUES (1); SELECT * FROM t LOCK IN SHARE MODE;"
a
1
```

The difference is that both spellings are row-locking clauses and are reported as unreadable SQL.

## Expected behavior

`LOCK IN SHARE MODE` after a table reference with no alias, and `FOR KEY SHARE`, are refused the way their siblings are: ErrUnsupportedSyntax carrying "a row-locking clause is not supported; SQLite locks the database rather than rows". LOCK stays a possible alias everywhere else, since it is not a reserved word in MySQL and a table can be aliased with it; it is the following IN that makes the clause. `FOR KEY SHARE` is one of PostgreSQL's four row-level lock strengths and belongs in the same list as the three already there.

## Cause (guess)

`parseRowLock` in dialect/internal/parser/from.go tests for `FOR UPDATE`, `FOR SHARE`, `FOR NO KEY` and `LOCK IN`, which misses `FOR KEY SHARE`, and the alias reader in the same file consumes LOCK as the table's alias before parseRowLock is reached, so by then the parser stands on IN.

## Tests to add

- In dialect/internal/parser/select_test.go, extend the table in TestQueryFormsThatAreNotImplemented with `SELECT a FROM t LOCK IN SHARE MODE` (MySQL, no alias) and `SELECT a FROM t FOR KEY SHARE` (PostgreSQL), asserting ErrUnsupportedSyntax.
- In the same file, assert the sibling that must keep working: `SELECT a FROM t LOCK` and `SELECT a FROM t AS lock_` still read LOCK as an alias, since MySQL allows it and only `LOCK IN` opens the clause.
- In dialect/dialect_test.go, assert the message a caller sees for both spellings names the row-locking clause rather than a token position, so the refusal text stays the one the other spellings give.

### LINKED ISSUE #919: A dump to a blank output directory writes into a directory named with spaces, and an empty one fails with a raw mkdir error
Found on main (c5fb518) while sweeping what this package says when it is handed a path it cannot use.

## Phenomenon

An output directory made only of spaces is taken as a directory name, so `filesql.DumpDatabase(db, "   ")` and `NewBuilder().AddPath(...).EnableAutoSave("  ")` create a directory whose name is those spaces in the process's working directory and write every table into it, reporting no error. The files land somewhere the caller did not name and cannot easily find, and this package refuses to read them back: `validatePath` in validator.go answers ErrEmptyPath for the same string, so `AddPath("  ")` cannot reopen what the dump just wrote. An empty output directory is the other half of the same hole: `DumpDatabase(db, "")` fails with the raw operating-system error `mkdir : no such file or directory` rather than the ErrEmptyPath this package has for exactly that, and `DumpDatabase` to a path that already exists as a regular file fails with the raw `mkdir /tmp/x/regular.txt: not a directory` where the auto-save path has a sentence of its own for that case.

## Reproduction

```go
package main

import (
	"context"
	"errors"
	"fmt"
	"os"
	"path/filepath"

	"github.com/nao1215/filesql"
)

func main() {
	dir, _ := os.MkdirTemp("", "probe")
	csv := filepath.Join(dir, "t.csv")
	os.WriteFile(csv, []byte("a,b\n1,2\n"), 0o600)
	work, _ := os.MkdirTemp("", "work")
	os.Chdir(work)

	db, _ := filesql.OpenContext(context.Background(), csv)
	defer db.Close()

	err := filesql.DumpDatabase(db, "")
	fmt.Printf("dump to \"\"    -> %v (ErrEmptyPath=%v)\n", err, errors.Is(err, filesql.ErrEmptyPath))
	fmt.Printf("dump to \"   \" -> %v\n", filesql.DumpDatabase(db, "   "))

	file := filepath.Join(dir, "regular.txt")
	os.WriteFile(file, []byte("x"), 0o600)
	fmt.Printf("dump onto a file -> %v\n", filesql.DumpDatabase(db, file))

	auto, _ := filesql.NewBuilder().AddPath(csv).EnableAutoSave("  ").Open(context.Background())
	auto.Close()

	entries, _ := os.ReadDir(work)
	for _, e := range entries {
		fmt.Printf("created in the working directory: %q dir=%v\n", e.Name(), e.IsDir())
	}
}
```

```
dump to ""    -> filesql: I/O operation failed: failed to create output directory: mkdir : no such file or directory (ErrEmptyPath=false)
dump to "   " -> <nil>
dump onto a file -> filesql: I/O operation failed: failed to create output directory: mkdir /tmp/probe.../regular.txt: not a directory
created in the working directory: "   " dir=true
created in the working directory: "  " dir=true
```

The difference is that a blank destination is written to instead of being refused, and that the two refusals that do happen are the operating system's words rather than this package's.

## Expected behavior

A destination that is empty or made only of whitespace is refused with ErrEmptyPath before anything is created, for `DumpDatabase`, `DumpDatabaseContext`, `EnableAutoSave` and `EnableAutoSaveOnCommit` alike; an auto-save configured that way is refused when the builder is opened, which is where the rest of the auto-save configuration is checked. The empty string keeps its one documented meaning in the auto-save calls, where EnableAutoSave("") is written down as "overwrite original files", and only a non-empty blank string is refused there. This is the reading validatePath already takes for an input path, and taking a different one for an output path is what lets a dump write a file no load can name. A destination that exists and is not a directory is refused with this package's own sentence, the one validateOutputDirectory already writes as "output path exists but is not a directory", rather than with a raw mkdir error; the package documents in internal/reader/delimited.go that a raw failure is replaced because it carries neither a sentinel nor a position.

## Cause (guess)

`dumpSQLiteDatabase` in save.go calls os.MkdirAll on the string it is handed with no check in front of it, so an empty string reaches mkdir and a blank one names a directory; `validateAutoSaveConfig` in validator.go tests `config.outputDir != ""` to decide whether to check the directory at all, so a whitespace string skips the check and later becomes a directory the same way.

## Tests to add

- In save_test.go, assert `DumpDatabase(db, "")`, `DumpDatabase(db, "   ")` and `DumpDatabase(db, "\t")` return ErrEmptyPath and that the working directory holds nothing afterwards, reading the directory before and after so a created directory fails the test.
- In save_test.go, assert a dump onto an existing regular file returns this package's error naming that the path is not a directory, matched with errors.Is on the sentinel it carries rather than on the operating system's text.
- In autosave_test.go, assert `EnableAutoSave("  ")` and `EnableAutoSaveOnCommit("  ")` are refused at Open with ErrEmptyPath, and the sibling that must keep working: `EnableAutoSave("")` still overwrites the original files.

nao1215:
Four bugs found by sweeping the spellings a dialect accepts for a construct this package already refuses or rewrites by name, plus the lint failure standing on main.

The worst of the four is silent: PostgreSQL writes ONLY in front of a table name to say a statement reaches that table and not the tables inheriting from it, and the parser read the keyword as the name and the name as its alias. `SELECT * FROM ONLY t` became `SELECT * FROM ONLY AS t`, and `UPDATE ONLY t SET a = 1` an update of a table called ONLY. A database with no such table answered "no such table: ONLY"; a database holding a file named only.csv answered from, updated or deleted the wrong table with nothing said. Nothing inherits from a table here, so ONLY now names the table beside it and is dropped, in all four spellings PostgreSQL accepts, as is the trailing star that is its opposite. MySQL 8.4 answers "Table 'db.ONLY' doesn't exist" for the same query, so MySQL and GoogleSQL keep reading the word as a table's name.

The other three are about what a caller is told. An ALTER TABLE with more than one change, a column placed with FIRST or AFTER, and the existence checks of ADD COLUMN, DROP COLUMN and ALTER TABLE were all refused with ErrInvalidSyntax, the sentinel documented as "the query could not be read" -- but each reads, and asks for what SQLite cannot make. They are now refused by name with ErrUnsupportedSyntax, where their siblings MODIFY, CHANGE and ADD CONSTRAINT already were, and CASCADE or RESTRICT after a dropped column is read and dropped, which is what DROP TABLE already does with those words. MySQL's LOCK IN SHARE MODE written straight after a table name was refused as unreadable because LOCK was taken for the alias, and PostgreSQL's FOR KEY SHARE was missing from the words that open a row lock; both now carry the refusal their siblings carry. And an export to a path of spaces created a directory of spaces in the working directory and wrote every table into it, which is a file no load can name, while an empty path failed with the operating system's own mkdir error.

Two things beside the fixes: staticcheck has been failing on main since the ISO-2022-JP work, because a test holds a raw U+001B in a string literal (ST1018); the literal is now written as an escape sequence, and the staticcheck configuration reports 0 issues locally. And a 6.2 MB scratch binary named `probe` had been committed by accident, so every module download carried it; it is deleted and ignored.

Verified with `make test`, `make lint`, the staticcheck configuration, a cross build for windows/amd64 and darwin/arm64, and sqly built and tested against this checkout through a local replace. The translate benchmarks are unchanged in allocations and within noise in time (mysql 13.5 to 14.6 microseconds median, postgresql and googlesql flat).

Closes #916
Closes #917
Closes #918
Closes #919

## Summary by CodeRabbit

- **Bug Fixes**
  - PostgreSQL `ONLY` table references now translate correctly.
  - Unsupported `ALTER TABLE` forms and row-locking clauses now return clear unsupported-syntax errors.
  - `CASCADE` and `RESTRICT` modifiers are handled correctly when dropping tables or columns.
  - Empty or whitespace-only export and auto-save paths are rejected before anything is created.
  - Invalid destination paths now return clearer, classifiable errors.
  - Empty-string auto-save destinations continue to support overwrite mode.

- **Documentation**
  - Updated SQL support and path-validation documentation.


nao1215:
All four review findings are fixed in the commit above. The star for PostgreSQL's inheritance spelling is no longer read when a dot stands in front of it, so `FROM s.*`, `UPDATE s.* SET ...` and `DELETE FROM s.*` are refused as before; a quoted identifier after ONLY is a name whatever word it spells, so `FROM ONLY \"from\"` names that table; the word that says what happens to what depends on a dropped table or column is read once rather than twice, so `CASCADE RESTRICT` in either order stays a refusal, for DROP TABLE as well as for a dropped column; and validateOutputDirectory now wraps both of its errors with ErrIOOperation, which is the sentinel DumpDatabase already used for the same condition. Each has a regression case: the refused spellings and the quoted names in dialect_test.go, and an auto-save onto an existing regular file in autosave_test.go.

PR #926, opened 2026-09-01
fix(dialect): read the value and the clause the caller wrote

### LINKED ISSUE #921: A DEFAULT written where a value goes becomes the text DEFAULT, so an UPDATE fills the column with a word
Found on main (c5fb518) while sweeping the spellings a dialect accepts for a construct this package already refuses or rewrites by name.

## Phenomenon

The DEFAULT keyword, which PostgreSQL and MySQL write where a value goes to mean "the column's own default", is read as a column name and rendered as a quoted identifier: `dialect.Translate(dialect.PostgreSQL, "UPDATE t SET a = DEFAULT")` answers `UPDATE t SET a = "DEFAULT"`. SQLite reads a double-quoted name that matches no column as a string literal, so the statement runs and writes the text `DEFAULT` into every row of the column. Nothing is reported at any point. A caller who writes the statement their engine takes gets their column filled with a word, and a database that happens to hold a column named DEFAULT -- a file whose header says so -- gets that column's value copied instead. The same reading applies inside an INSERT: `INSERT INTO t (a, b) VALUES (DEFAULT, 1)` becomes `INSERT INTO t (a, b) VALUES ("DEFAULT", 1)`.

## Reproduction

```go
package main

import (
	"database/sql"
	"fmt"

	_ "modernc.org/sqlite"
)

func main() {
	out, err := dialect.Translate(dialect.PostgreSQL, "UPDATE t SET a = DEFAULT")
	fmt.Printf("translated: %q err=%v\n", out, err)

	defer db.Close()
	db.SetMaxOpenConns(1)
	db.Exec("CREATE TABLE t (a TEXT DEFAULT 'd', b INTEGER)")
	db.Exec("INSERT INTO t VALUES ('x', 1)")
	if _, err := db.Exec(out); err != nil {
		fmt.Println("sqlite:", err)
	}
	var a string
	db.QueryRow("SELECT a FROM t").Scan(&a)
	fmt.Printf("a is now %q\n", a)
}
```

```
translated: "UPDATE t SET a = \"DEFAULT\"" err=<nil>
a is now "DEFAULT"
```

The row held `x` and the column's default is `d`; the statement wrote the word `DEFAULT` and reported nothing.

## Expected behavior

DEFAULT standing where a value goes is refused with ErrUnsupportedSyntax, naming the keyword: SQLite has no way to write a column's default into a row, so there is nothing to translate it into, and a refusal is the answer this package gives elsewhere for a construct SQLite cannot express. The three positions are `UPDATE ... SET col = DEFAULT`, a DEFAULT among the values of an INSERT, and DEFAULT standing alone as an expression, which is not valid in the source dialects either and must not become a quoted name. `INSERT INTO t DEFAULT VALUES` is a different construct that SQLite does take, and it already translates; it must keep working. A column that really is named DEFAULT is still reachable by quoting it, which is what a caller writes in their own dialect too.

## Cause (guess)

The expression parser in dialect/internal/parser/primary.go reads any word that is not one of the words it knows as a name, and DEFAULT is not among them where a value is expected, so it becomes an ast.Ident and the renderer quotes it as an identifier. The word already appears in the parser's clauseKeywords table, which is about aliases rather than about values.

## Tests to add

- In dialect/dialect_test.go, assert `UPDATE t SET a = DEFAULT`, `UPDATE t SET a = DEFAULT WHERE b = 1`, `INSERT INTO t (a) VALUES (DEFAULT)` and `INSERT INTO t (a, b) VALUES (DEFAULT, 1)` are refused with ErrUnsupportedSyntax in MySQL and PostgreSQL, and that the message names DEFAULT.
- In the same file, assert the siblings that must keep working: `INSERT INTO t DEFAULT VALUES` still translates, and a column named DEFAULT written as a quoted identifier still reads as a column.
- Add a case to the test that prepares translated SQL against SQLite so no translation of these statements can silently become a string literal again: a rendered `"DEFAULT"` where a value belongs is exactly what SQLite accepts and this package must not produce.

### LINKED ISSUE #922: GoogleSQL THEN RETURN and PostgreSQL OVERRIDING SYSTEM VALUE are refused as unreadable SQL
Found on main (c5fb518) while sweeping the spellings a dialect accepts for a construct this package already refuses or rewrites by name.

## Phenomenon

Two clauses that are valid where they are written come back as ErrInvalidSyntax, the sentinel this package documents for a query that could not be read. GoogleSQL's `THEN RETURN`, which is how Cloud Spanner spells the returning clause of a DML statement, is refused at the keyword; PostgreSQL's `OVERRIDING SYSTEM VALUE` in an INSERT is refused the same way. The returning clause is the one this package already translates for PostgreSQL and renders as SQLite's RETURNING, so the same request in the other spelling is reported as a typing mistake instead.

## Reproduction

```go
package main

import (
	"fmt"

)

func main() {
	for _, c := range []struct {
		d dialect.Dialect
		q string
	}{
		{dialect.GoogleSQL, "INSERT INTO t (a) VALUES (1) THEN RETURN a"},
		{dialect.GoogleSQL, "DELETE FROM t WHERE a = 1 THEN RETURN a"},
		{dialect.GoogleSQL, "UPDATE t SET a = 1 WHERE b = 2 THEN RETURN *"},
		{dialect.PostgreSQL, "INSERT INTO t (a) OVERRIDING SYSTEM VALUE VALUES (1)"},
		{dialect.PostgreSQL, "INSERT INTO t (a) VALUES (1) RETURNING a"},
	} {
		out, err := dialect.Translate(c.d, c.q)
		fmt.Printf("%-11v %-54s -> %q %v\n", c.d, c.q, out, err)
	}
}
```

```
googlesql   INSERT INTO t (a) VALUES (1) THEN RETURN a             -> "" dialect: invalid SQL syntax: expected end of statement, found THEN at line 1, column 30
googlesql   DELETE FROM t WHERE a = 1 THEN RETURN a                -> "" dialect: invalid SQL syntax: expected end of statement, found THEN at line 1, column 27
googlesql   UPDATE t SET a = 1 WHERE b = 2 THEN RETURN *           -> "" dialect: invalid SQL syntax: expected end of statement, found THEN at line 1, column 31
postgresql  INSERT INTO t (a) OVERRIDING SYSTEM VALUE VALUES (1)   -> "" dialect: invalid SQL syntax: expected end of statement, found OVERRIDING at line 1, column 19
postgresql  INSERT INTO t (a) VALUES (1) RETURNING a               -> "INSERT INTO t (a) VALUES (1) RETURNING a" <nil>
```

The difference is that the last line shows the same clause, in the spelling this package reads, translating without trouble.

## Expected behavior

`THEN RETURN` is read as the returning clause it is and lowered onto SQLite's RETURNING, so `INSERT INTO t (a) VALUES (1) THEN RETURN a` translates to `INSERT INTO t (a) VALUES (1) RETURNING a`, and the naming rule the package already applies to a RETURNING list applies to it unchanged. It is the whole of the difference between the two spellings, and the tree already models the clause. `OVERRIDING SYSTEM VALUE` says what to do about an identity column, which a table here does not have, so it belongs with the words that are read and dropped -- as PostgreSQL's UNLOGGED and CONCURRENTLY already are -- or, if it is refused, refused by name with ErrUnsupportedSyntax rather than reported as unreadable.

## Cause (guess)

The statement parsers in dialect/internal/parser/stmt.go read RETURNING by that word alone, and nothing looks for THEN RETURN before the end of the statement is expected; the INSERT parser expects VALUES or a query straight after the column list, so the words between them stop it.

## Tests to add

- In dialect/dialect_test.go, assert the three GoogleSQL statements above translate and that their output carries RETURNING with the columns the caller named, including the star form.
- In the same file, assert the naming rule holds for the new spelling: a `THEN RETURN` list of an expression that lowering rewrites keeps the caller's own text as the column name, which is what the existing RETURNING test pins.
- Assert `OVERRIDING SYSTEM VALUE` and `OVERRIDING USER VALUE` no longer answer ErrInvalidSyntax, and that an INSERT written with either still names the columns and values the caller wrote.

### LINKED ISSUE #923: GROUP BY DISTINCT, COMMIT AND CHAIN, INHERITS and Spanner's trailing PRIMARY KEY are refused as unreadable SQL
Found on main (c5fb518) while sweeping the spellings a dialect accepts for a construct this package already refuses or rewrites by name. This is the same class as the ALTER TABLE spellings, in the statements beside it.

## Phenomenon

Four statements their own engine accepts come back as ErrInvalidSyntax, the sentinel documented for a query that could not be read, and a fifth is refused with a diagnosis about something the caller never wrote. PostgreSQL's `GROUP BY DISTINCT a`, its `COMMIT AND CHAIN`, its `CREATE TABLE t (a INT) INHERITS (u)` and GoogleSQL's `CREATE TABLE t (a INT64) PRIMARY KEY (a)`, which is how Cloud Spanner declares a primary key, all stop the parser at a token. `CREATE TABLE t (LIKE u)` is refused with "the column type U is not supported", although MySQL's spelling of that same statement, `CREATE TABLE t LIKE u`, is already refused by name with "CREATE TABLE ... LIKE is not supported; SQLite has no statement that copies a table's structure". A caller who branches on the sentinel to tell a typing mistake from a construct this package does not carry takes the wrong branch, and the one who reads the message is sent to look at a column type that is not in their statement.

## Reproduction

```go
package main

import (
	"fmt"

)

func main() {
	for _, c := range []struct {
		d dialect.Dialect
		q string
	}{
		{dialect.PostgreSQL, "SELECT a FROM t GROUP BY DISTINCT a"},
		{dialect.PostgreSQL, "COMMIT AND CHAIN"},
		{dialect.PostgreSQL, "COMMIT AND NO CHAIN"},
		{dialect.PostgreSQL, "CREATE TABLE t (a INT) INHERITS (u)"},
		{dialect.GoogleSQL, "CREATE TABLE t (a INT64) PRIMARY KEY (a)"},
		{dialect.PostgreSQL, "CREATE TABLE t (LIKE u)"},
		{dialect.MySQL, "CREATE TABLE t LIKE u"},
	} {
		_, err := dialect.Translate(c.d, c.q)
		fmt.Printf("%-11v %-42s -> %v\n", c.d, c.q, err)
	}
}
```

```
postgresql  SELECT a FROM t GROUP BY DISTINCT a        -> dialect: invalid SQL syntax: expected end of statement, found a at line 1, column 35
postgresql  COMMIT AND CHAIN                           -> dialect: invalid SQL syntax: expected end of statement, found AND at line 1, column 8
postgresql  COMMIT AND NO CHAIN                        -> dialect: invalid SQL syntax: expected end of statement, found AND at line 1, column 8
postgresql  CREATE TABLE t (a INT) INHERITS (u)        -> dialect: invalid SQL syntax: expected end of statement, found ( line 1, column 33
googlesql   CREATE TABLE t (a INT64) PRIMARY KEY (a)   -> dialect: invalid SQL syntax: expected end of statement, found ( at line 1, column 38
postgresql  CREATE TABLE t (LIKE u)                    -> dialect: syntax not supported on SQLite backend: the column type U is not supported; SQLite stores values as INTEGER, REAL, TEXT, BLOB or NUMERIC
mysql       CREATE TABLE t LIKE u                      -> dialect: syntax not supported on SQLite backend: CREATE TABLE ... LIKE is not supported; SQLite has no statement that copies a table's structure
```

PostgreSQL 17 runs the first four of these:

```
$ docker run --rm -d --name pg -e POSTGRES_PASSWORD=pw -e POSTGRES_DB=probe postgres:17-alpine
$ docker exec pg psql -U postgres -d probe -c "CREATE TABLE u(a int); CREATE TABLE t2(a int) INHERITS (u); CREATE TABLE t3 (LIKE u);" -c "SELECT a FROM u GROUP BY DISTINCT a;" -c "BEGIN;" -c "COMMIT AND CHAIN;" -c "COMMIT AND NO CHAIN;"
```

The difference is that each reads in its own dialect and asks for something SQLite does not do, which is a different answer from "could not be read".

## Expected behavior

Each spelling is read to its end and answered by name. `GROUP BY DISTINCT` deduplicates the grouping sets, and with a plain list of expressions it groups exactly as `GROUP BY` does, so the word can be read and dropped there and refused where grouping sets stand behind it. `AND CHAIN` starts the next transaction as the current one ends, which SQLite has no form for, so it is refused with ErrUnsupportedSyntax naming it; `AND NO CHAIN` is what a plain COMMIT already does and is read and dropped. `INHERITS` declares a table hierarchy SQLite does not have and is refused by name. Spanner's trailing `PRIMARY KEY (a)` is the same primary key SQLite writes inside the table body, so it belongs there -- `CREATE TABLE t (a INT64) PRIMARY KEY (a)` translating to `CREATE TABLE t (a INTEGER, PRIMARY KEY (a))` -- and if it is refused instead it is refused by name. `CREATE TABLE t (LIKE u)` is the statement MySQL spells `CREATE TABLE t LIKE u` and deserves that same message rather than a report about a column type.

## Cause (guess)

Each of these ends a statement in dialect/internal/parser: parseGroupBy reads the terms with nothing looking for DISTINCT or ALL in front of them, the transaction parser in stmt.go stops after COMMIT, parseTableOptions in ddl.go does not know INHERITS, and the CREATE TABLE parser expects the statement to end after the body, so Spanner's primary key stops it. The LIKE inside a table body is read as a column named LIKE with the type U, which is the shape the ADD CONSTRAINT misdiagnosis had before it was fixed.

## Tests to add

- In dialect/dialect_test.go, assert `COMMIT AND CHAIN`, `CREATE TABLE t (a INT) INHERITS (u)` and `CREATE TABLE t (LIKE u)` answer ErrUnsupportedSyntax with a message naming the construct, and that none of the five answers ErrInvalidSyntax.
- In the same file, assert what is read and dropped keeps working: `COMMIT AND NO CHAIN` translates to `COMMIT`, and `SELECT a FROM t GROUP BY DISTINCT a` translates to the same SQL as `GROUP BY a`.
- Assert `CREATE TABLE t (LIKE u)` and `CREATE TABLE t LIKE u` carry the same message, so the two spellings of one statement cannot drift apart again.
- Whichever way Spanner's trailing primary key is answered, pin it: either the translation puts the key inside the table body, in which case assert the rendered SQL and that SQLite prepares it, or the refusal names it.

### LINKED ISSUE #924: A PostgreSQL IS JSON predicate reaches SQLite as a comparison against a column named JSON
Found on main (c5fb518) while sweeping the spellings a dialect accepts for a construct this package already refuses or rewrites by name.

## Phenomenon

PostgreSQL's JSON predicate, `a IS JSON`, is passed through to SQLite untranslated. SQLite has no such predicate: it reads IS as its null-safe comparison and JSON as a name, so the query becomes a comparison between the column and whatever JSON names. On a database that holds a column called json -- a file whose header says so, which is an ordinary thing for a CSV to say -- the query answers that comparison and reports nothing, where the caller asked which rows hold valid JSON. On a database without one it fails with "no such column: JSON", which names something the caller did not write. The package documents that a construct it does not implement is refused "rather than forwarded to SQLite untranslated", and this one is forwarded.

## Reproduction

```go
package main

import (
	"database/sql"
	"fmt"

	_ "modernc.org/sqlite"
)

func main() {
	out, err := dialect.Translate(dialect.PostgreSQL, "SELECT a FROM t WHERE a IS JSON")
	fmt.Printf("translated: %q err=%v\n", out, err)

	defer db.Close()
	db.SetMaxOpenConns(1)
	db.Exec("CREATE TABLE t (a TEXT, json TEXT)")
	db.Exec("INSERT INTO t VALUES ('{}', 'x')")
	rows, err := db.Query(out)
	if err != nil {
		fmt.Println("sqlite:", err)
		return
	}
	n := 0
	for rows.Next() {
		n++
	}
	rows.Close()
	fmt.Printf("%d rows, and the one row holds valid JSON\n", n)
}
```

```
translated: "SELECT a FROM t WHERE a IS JSON" err=<nil>
0 rows, and the one row holds valid JSON
```

The difference is that the predicate asks whether the value is JSON and the translation asks whether it equals a column.

## Expected behavior

SQLite carries `json_valid(X)`, which answers exactly what `X IS JSON` asks, so the predicate translates to it: `a IS JSON` becomes `json_valid(a)` and `a IS NOT JSON` becomes `NOT json_valid(a)`, with the select-item naming rule this package already applies to a rewritten expression keeping the caller's own text as the column name. The forms that narrow the question -- `IS JSON OBJECT`, `IS JSON ARRAY`, `IS JSON SCALAR`, `IS JSON VALUE` -- are answered by `json_type(X)`, and `WITH UNIQUE KEYS` has no SQLite form and is refused by name with ErrUnsupportedSyntax. What must not happen in any of these is what happens now, which is that a predicate reaches SQLite as a comparison against a name.

## Cause (guess)

The IS operator in dialect/internal/parser/expr.go reads the word after IS as an operand when it is not one of NULL, TRUE, FALSE, DISTINCT or UNKNOWN, so JSON becomes an identifier and the renderer writes it back as one.

## Tests to add

- In dialect/dialect_test.go, assert `SELECT a FROM t WHERE a IS JSON` and `... IS NOT JSON` translate to the json_valid form, and that the translated query answers the rows that hold valid JSON when it is prepared and run against SQLite, which is the check a pass-through cannot pass.
- Assert the narrowed forms: `IS JSON OBJECT`, `IS JSON ARRAY` and `IS JSON SCALAR` either translate through json_type or are refused by name, and that `IS JSON WITH UNIQUE KEYS` is ErrUnsupportedSyntax rather than ErrInvalidSyntax.
- Assert the sibling that must not change: a column really named json is still readable, so `SELECT json FROM t` and `WHERE a IS json` written in lower case with a column of that name keep their meaning.

### LINKED ISSUE #925: AT TIME ZONE, a time-zone typed literal and OPERATOR(schema.op) are refused as unreadable SQL
Found on main (c5fb518) while sweeping the spellings a dialect accepts for a construct this package already refuses or rewrites by name.

## Phenomenon

The time-zone forms of PostgreSQL and GoogleSQL are refused with ErrInvalidSyntax, the sentinel this package documents for a query that could not be read: `SELECT a AT TIME ZONE 'UTC' FROM t`, the same clause inside an EXTRACT, and the typed literal `TIMESTAMP WITH TIME ZONE '2024-01-01 00:00:00+00'`. PostgreSQL's `OPERATOR(pg_catalog.+)`, which names an operator by its schema, is refused the same way. Each reads in its own dialect, and the package's own timestamp work translates everything around them -- `a - INTERVAL '1 day'`, `DATE_ADD(a, INTERVAL 1 DAY)`, `TIMESTAMP_TRUNC(a, DAY)`, `a::timestamptz` -- so the caller is told their SQL is unreadable for the one clause in the middle of statements this package otherwise carries.

## Reproduction

```go
package main

import (
	"fmt"

)

func main() {
	for _, c := range []struct {
		d dialect.Dialect
		q string
	}{
		{dialect.PostgreSQL, "SELECT a AT TIME ZONE 'UTC' FROM t"},
		{dialect.PostgreSQL, "SELECT to_timestamp(0) AT TIME ZONE 'UTC'"},
		{dialect.GoogleSQL, "SELECT EXTRACT(DAY FROM a AT TIME ZONE 'UTC') FROM t"},
		{dialect.PostgreSQL, "SELECT TIMESTAMP WITH TIME ZONE '2024-01-01 00:00:00+00'"},
		{dialect.PostgreSQL, "SELECT a OPERATOR(pg_catalog.+) b FROM t"},
		{dialect.PostgreSQL, "SELECT a - INTERVAL '1 day' FROM t"},
	}
		{
		_, err := dialect.Translate(c.d, c.q)
		fmt.Printf("%-11v %-56s -> %v\n", c.d, c.q, err)
	}
}
```

```
postgresql  SELECT a AT TIME ZONE 'UTC' FROM t                       -> dialect: invalid SQL syntax: expected end of statement, found TIME at line 1, column 13
postgresql  SELECT to_timestamp(0) AT TIME ZONE 'UTC'                -> dialect: invalid SQL syntax: expected end of statement, found TIME at line 1, column 27
googlesql   SELECT EXTRACT(DAY FROM a AT TIME ZONE 'UTC') FROM t     -> dialect: invalid SQL syntax: expected ), found AT at line 1, column 27
postgresql  SELECT TIMESTAMP WITH TIME ZONE '2024-01-01 00:00:00+00' -> dialect: invalid SQL syntax: expected end of statement, found WITH at line 1, column 18
postgresql  SELECT a OPERATOR(pg_catalog.+) b FROM t                 -> dialect: invalid SQL syntax: expected end of statement, found ( at line 1, column 18
postgresql  SELECT a - INTERVAL '1 day' FROM t                       -> translated
```

The last line is the same statement family translating, which is what makes the refusal above it read as a report about the caller's typing.

## Expected behavior

Each is read to its end and answered by name. `AT TIME ZONE` moves a timestamp between zones, which SQLite can express for the UTC case its datetime functions carry and cannot for an arbitrary zone name, so it is either lowered onto those functions where that is exact or refused with ErrUnsupportedSyntax saying SQLite keeps no zone with a timestamp -- and in the second case the refusal must name the clause rather than the token after it. The typed literal `TIMESTAMP WITH TIME ZONE 'text'` is the same decision in the type's spelling and follows it; `TIMESTAMP WITHOUT TIME ZONE 'text'` is the plain timestamp this package already carries and should translate. `OPERATOR(schema.op)` names an operator by its schema and the operators it can name are the ones already translated, so it is read and the operator inside it used, or refused by name.

## Cause (guess)

The expression parser in dialect/internal/parser/expr.go has no postfix for AT, so the words after an operand end the expression and the statement is expected to end there; the typed-literal reader in primary.go matches TIMESTAMP followed by a string and does not look for WITH TIME ZONE between them; OPERATOR is read as a name, so the parenthesis after it opens what looks like a call in a place a call cannot stand.

## Tests to add

- In dialect/dialect_test.go, assert none of the five statements answers ErrInvalidSyntax, and pin whichever answer is chosen for each: the rendered SQL when it translates, or ErrUnsupportedSyntax with a message naming the clause when it does not.
- Assert the sibling spelling that must translate either way: `TIMESTAMP WITHOUT TIME ZONE '2024-01-01 00:00:00'` and `TIMESTAMP '2024-01-01 00:00:00'` answer the same SQL.
- Add the EXTRACT case, since the clause inside a keyword-separated call is read by a different path than the one in a select list, and a fix that only reaches the select list would leave it refused.

nao1215:
Five more bugs from the same sweep as the previous pull request: write a construct in every spelling its dialect accepts, and see where the answer changes.

Two of them are silent. DEFAULT written where a value goes -- `UPDATE t SET a = DEFAULT`, `INSERT INTO t (a, b) VALUES (DEFAULT, 1)` -- was read as a column name and rendered as the quoted identifier `"DEFAULT"`; SQLite reads a quoted name that matches no column as a string, so the statement ran and wrote the word DEFAULT into the column with nothing reported. PostgreSQL's `a IS JSON` reached SQLite untranslated, where IS is the null-safe comparison and JSON is a name, so the predicate asked whether the value equalled a column called JSON: a file whose header says json is an ordinary thing to load, and the query then answered that comparison in silence. SQLite has no way to write a column's default into a row, so DEFAULT is refused by name; it does carry `json_valid`, which asks exactly what the predicate asks, so IS JSON becomes that.

The rest is the classification fault the previous pull request fixed for ALTER TABLE, in the clauses beside it: nine spellings an engine accepts answered with ErrInvalidSyntax, the sentinel documented for a query that could not be read. Four are now read and answered -- Cloud Spanner's `THEN RETURN` becomes the RETURNING it is and its trailing `PRIMARY KEY (a)` becomes the table constraint SQLite takes, while `OVERRIDING SYSTEM VALUE`, the ALL or DISTINCT before PostgreSQL's grouping elements and `AND NO CHAIN` are read and dropped because none of them changes the rows. Five are refused by name: `AT TIME ZONE`, a timestamp typed `WITH TIME ZONE`, `OPERATOR(schema.op)`, `AND CHAIN` and `INHERITS`. And `CREATE TABLE t (LIKE u)` was refused with "the column type U is not supported", a report about something the caller never wrote; it now carries the sentence MySQL's spelling of that same statement already had.

The four PostgreSQL statements were run against postgres:17-alpine to confirm each is a statement that engine takes.

Verified with `make test`, `make lint`, the staticcheck configuration, a cross build for windows/amd64, and sqly built and tested against this checkout through a local replace. The translate benchmarks are unchanged in allocations and within noise in time.

Closes #921
Closes #922
Closes #923
Closes #924
Closes #925

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of SQL dialect-specific syntax and clearer errors for unsupported constructs.
  * Added translation for JSON predicates and alternate `RETURNING` syntax.
  * Correctly handles trailing primary key declarations and selected grouping and transaction clauses.
  * Prevents keywords such as `DEFAULT` from being misread as column names.
  * Improved handling of time-zone syntax, table inheritance, table copying, and related unsupported clauses.

* **Documentation**
  * Expanded documentation covering supported translations and unsupported SQL syntax.


nao1215:
Both review findings are fixed in the commit above. The ALL or DISTINCT before PostgreSQL's grouping elements is now read whatever follows it, so `GROUP BY DISTINCT 1` and `GROUP BY ALL (a + b)` translate and a modifier with no grouping element after it stays the syntax error it is; OVERRIDING is gated to PostgreSQL, so MySQL and GoogleSQL refuse a statement neither of them takes. Both have a case in dialect_test.go.

PR #930, opened 2026-09-01
fix: name the thing the caller named

### LINKED ISSUE #927: ANALYZE VERBOSE t and ANALYZE TABLE t analyze the option word and drop the table the caller named
Found on main (a0f5250) while sweeping the spellings a dialect accepts for a statement this package already translates.

## Phenomenon

An ANALYZE written the way its own dialect writes it names the wrong object. PostgreSQL's `ANALYZE VERBOSE t` translates to `ANALYZE VERBOSE`: the option is read as the table's name and the table the caller named is dropped, so the statement asks SQLite to analyze something called VERBOSE. MySQL's `ANALYZE TABLE t` translates to `ANALYZE "TABLE"` for the same reason. Neither reports anything; SQLite answers "no such table: VERBOSE" for one and, on a database that holds a table of that name, analyzes it instead of the one the caller wrote. The parenthesized option list PostgreSQL also accepts, `ANALYZE (VERBOSE) t`, and MySQL's `ANALYZE NO_WRITE_TO_BINLOG TABLE t` are refused with ErrInvalidSyntax, which is the sentinel for a query that could not be read.

## Reproduction

```go
package main

import (
	"fmt"

)

func main() {
	for _, c := range []struct {
		d dialect.Dialect
		q string
	}{
		{dialect.PostgreSQL, "ANALYZE t"},
		{dialect.PostgreSQL, "ANALYZE VERBOSE t"},
		{dialect.PostgreSQL, "ANALYZE (VERBOSE) t"},
		{dialect.MySQL, "ANALYZE TABLE t"},
		{dialect.MySQL, "ANALYZE NO_WRITE_TO_BINLOG TABLE t"},
	} {
		out, err := dialect.Translate(c.d, c.q)
		fmt.Printf("%-11v %-36s -> %-20q %v\n", c.d, c.q, out, err)
	}
}
```

```
postgresql  ANALYZE t                            -> "ANALYZE t"          <nil>
postgresql  ANALYZE VERBOSE t                    -> "ANALYZE VERBOSE"    <nil>
postgresql  ANALYZE (VERBOSE) t                  -> ""                   dialect: invalid SQL syntax: expected end of statement, found ( at line 1, column 9
mysql       ANALYZE TABLE t                      -> "ANALYZE \"TABLE\""  <nil>
mysql       ANALYZE NO_WRITE_TO_BINLOG TABLE t   -> ""                   dialect: invalid SQL syntax: expected end of statement, found TABLE at line 1, column 28
```

The difference is that the second and fourth statements name t and the translations do not.

## Expected behavior

The words that are not the object's name are read and dropped, and the name is kept: `ANALYZE VERBOSE t`, `ANALYZE (VERBOSE) t`, `ANALYZE (VERBOSE, SKIP_LOCKED) t`, `ANALYZE TABLE t` and `ANALYZE NO_WRITE_TO_BINLOG TABLE t` all translate to `ANALYZE t`, since none of those words changes what is analyzed and SQLite has no form for any of them. `ANALYZE` with nothing after it stays what it is, which is every table. A statement that names no table must not come out naming one: this package's own rule is that a translation answers the question the caller asked, and analyzing a different object is the same class of fault as answering from a different table.

## Cause (guess)

parseAnalyze in dialect/internal/parser/stmt.go reads one name after the word ANALYZE and stops, so the first word it meets becomes the name whether it is one or not, and a parenthesis where it expects a name ends the statement.

## Tests to add

- In dialect/dialect_test.go, assert the five spellings above all translate to `ANALYZE t`, and that `ANALYZE` alone still translates to `ANALYZE`.
- Assert the sibling that must keep working: a table really named verbose is still analyzable when it is written as the name, `ANALYZE verbose` staying `ANALYZE verbose` in a dialect where VERBOSE is not an option word, and quoted otherwise.
- Add a case to the test that prepares translated SQL against SQLite, with a real table, so a translation that names an object the schema does not have fails there.

### LINKED ISSUE #928: CREATE INDEX INCLUDE, WITH CHECK OPTION and SELECT INTO are refused as unreadable SQL
Found on main (a0f5250) while sweeping the spellings a dialect accepts for a construct this package already reads or refuses by name.

## Phenomenon

Five more clauses valid where they are written come back with ErrInvalidSyntax, the sentinel documented for a query that could not be read: PostgreSQL's `CREATE INDEX i ON t (a) INCLUDE (b)`, its `CREATE VIEW v AS SELECT 1 WITH CHECK OPTION`, its `SELECT a INTO newt FROM t`, and MySQL's `SELECT a INTO OUTFILE '/tmp/x' FROM t` and `SELECT a INTO @v FROM t`. The clauses around them are already read: this package drops PostgreSQL's CONCURRENTLY and its `USING btree` on the same CREATE INDEX and translates the WHERE of a partial index, so the caller is told their statement is unreadable for the one clause in the middle of a statement it otherwise carries.

## Reproduction

```go
package main

import (
	"fmt"

)

func main() {
	for _, c := range []struct {
		d dialect.Dialect
		q string
	}{
		{dialect.PostgreSQL, "CREATE INDEX i ON t (a) INCLUDE (b)"},
		{dialect.PostgreSQL, "CREATE INDEX CONCURRENTLY i ON t USING btree (a) WHERE a > 1"},
		{dialect.PostgreSQL, "CREATE VIEW v AS SELECT 1 WITH CHECK OPTION"},
		{dialect.PostgreSQL, "SELECT a INTO newt FROM t"},
		{dialect.MySQL, "SELECT a INTO OUTFILE '/tmp/x' FROM t"},
		{dialect.MySQL, "SELECT a INTO @v FROM t"},
	} {
		out, err := dialect.Translate(c.d, c.q)
		fmt.Printf("%-11v %-58s -> %-38q %v\n", c.d, c.q, out, err)
	}
}
```

```
postgresql  CREATE INDEX i ON t (a) INCLUDE (b)                        -> ""                                     dialect: invalid SQL syntax: expected end of statement, found INCLUDE at line 1, column 25
postgresql  CREATE INDEX CONCURRENTLY i ON t USING btree (a) WHERE a > 1 -> "CREATE INDEX i ON t (a) WHERE a > 1"   <nil>
postgresql  CREATE VIEW v AS SELECT 1 WITH CHECK OPTION                -> ""                                     dialect: invalid SQL syntax: expected end of statement, found WITH at line 1, column 27
postgresql  SELECT a INTO newt FROM t                                  -> ""                                     dialect: invalid SQL syntax: expected end of statement, found INTO at line 1, column 10
mysql       SELECT a INTO OUTFILE '/tmp/x' FROM t                      -> ""                                     dialect: invalid SQL syntax: expected end of statement, found INTO at line 1, column 10
mysql       SELECT a INTO @v FROM t                                    -> ""                                     dialect: invalid SQL syntax: expected end of statement, found INTO at line 1, column 10
```

The second line is the same statement with two other clauses this package already reads and drops, which is what makes the first line read as a report about the caller's typing.

## Expected behavior

INCLUDE names columns an index carries without keying on them, which changes how the index is stored and not what a query answers, so it belongs with CONCURRENTLY and `USING btree` and is read and dropped. `WITH CHECK OPTION` constrains writes made through a view, and SQLite refuses every write through a view that has no INSTEAD OF trigger, so the clause can never bind: it is read and dropped, with the reason written down where the other dropped clauses are. PostgreSQL's `SELECT ... INTO name` is the statement SQLite spells `CREATE TABLE name AS SELECT ...`, so it translates to that; MySQL's `INTO OUTFILE` writes a file on the server and `INTO @var` fills a session variable, neither of which SQLite has, so both are refused by name with ErrUnsupportedSyntax rather than reported as unreadable.

## Cause (guess)

parseCreateIndex and parseCreateView in dialect/internal/parser/ddl.go stop after the parts they model, and the select parser in select.go has no INTO among the clauses it reads, so each leftover word is reported where the statement was expected to end.

## Tests to add

- In dialect/dialect_test.go, assert `CREATE INDEX i ON t (a) INCLUDE (b)` translates to the same SQL as `CREATE INDEX i ON t (a)`, and `CREATE VIEW v AS SELECT 1 WITH CHECK OPTION` to the same as the view without it.
- Assert `SELECT a INTO newt FROM t` translates to `CREATE TABLE newt AS SELECT a FROM t`, and that the translated statement is one SQLite prepares.
- Assert MySQL's two INTO spellings answer ErrUnsupportedSyntax with a message naming the clause, and that none of the five answers ErrInvalidSyntax.

### LINKED ISSUE #929: A file whose compression suffix does not match its bytes answers ErrCompression for three codecs and ErrParsing for five
Found on main (a0f5250) while comparing what this package says for one fault across the codecs it reads.

## Phenomenon

A file whose compression suffix does not match its bytes is reported two different ways depending on the codec. `data.csv.gz`, `.xz` and `.z` answer ErrCompression with "failed to create gzip reader: gzip: invalid header", which names the fault; `.bz2`, `.zst`, `.lz4`, `.s2` and `.snappy` answer ErrParsing with "failed to read CSV header: s2: corrupt input", which names the CSV reader although the CSV was never reached. The difference is not in the file or in the caller's request: it is that three of those decoders read their header when they are created and the rest fail on the first read, which happens inside the CSV header parse. A caller who branches on ErrCompression to tell "this file is not really compressed" from "this file is not really a CSV" gets the right answer for three codecs out of eight, and the message they show a user points at the wrong half of the file for the other five.

## Reproduction

```go
package main

import (
	"context"
	"errors"
	"fmt"
	"os"
	"path/filepath"

	"github.com/nao1215/filesql"
)

func main() {
	dir, _ := os.MkdirTemp("", "probe")
	for _, ext := range []string{"gz", "bz2", "xz", "zst", "lz4", "s2", "z", "snappy"} {
		path := filepath.Join(dir, "t.csv."+ext)
		os.WriteFile(path, []byte("id,name\n1,alice\n"), 0o600)
		_, err := filesql.OpenContext(context.Background(), path)
		fmt.Printf("%-7s compression=%-6v parsing=%-6v %v\n",
			ext, errors.Is(err, filesql.ErrCompression), errors.Is(err, filesql.ErrParsing), err)
	}
}
```

```
gz      compression=true   parsing=true   filesql: compression operation failed: failed to create gzip reader: gzip: invalid header
bz2     compression=false  parsing=true   filesql: parsing failed: failed to read CSV header: bzip2 data invalid: bad magic value
xz      compression=true   parsing=true   filesql: compression operation failed: failed to create xz reader: xz: invalid header magic bytes
zst     compression=false  parsing=true   filesql: parsing failed: failed to read CSV header: invalid input: magic number mismatch
lz4     compression=false  parsing=true   filesql: parsing failed: failed to read CSV header: newState: lz4: bad magic number
s2      compression=false  parsing=true   filesql: parsing failed: failed to read CSV header: s2: corrupt input
z       compression=true   parsing=true   filesql: compression operation failed: failed to create zlib reader: zlib: invalid header
snappy  compression=false  parsing=true   filesql: parsing failed: failed to read CSV header: s2: corrupt input
```

The difference is that one fault carries two sentinels and two sentences depending on which decoder happens to check its header early.

## Expected behavior

A file that cannot be decompressed answers ErrCompression whichever codec it claims, with a message naming the compression rather than the CSV header, and a file that decompresses into something that is not the format it claims answers ErrParsing. The eight codecs this package reads are one contract, and the repository already holds that reading: the codec tests assert every codec answers the same table for the same bytes, and the reader documents that a raw failure is replaced because it carries neither a sentinel nor a position. The fix is on this side of the dependency line: the decoders that fail lazily do so on the first read, so the read that fills the format reader's first bytes is where the failure has to be recognized and wrapped, rather than being passed on as whatever the format reader made of it.

## Cause (guess)

internal/codec wraps a reader per codec, and gzip, xz and zlib validate their header inside the constructor, which the loader already wraps with ErrCompression; bzip2, zstd, lz4, s2 and snappy return a reader that fails on the first Read, and that read happens inside internal/reader while it is parsing the format's header, where the error is wrapped with ErrParsing and the sentence about a CSV header.

## Tests to add

- In compression_test.go, assert every codec answers ErrCompression, and not the CSV sentence, for a file whose suffix says it is compressed and whose bytes are plain text; drive the list from the same table of codecs the equivalence test uses, so a codec added later is covered without a new case.
- Assert the sibling that must keep its own answer: a correctly compressed file holding text that is not the claimed format still answers ErrParsing, naming the format's reader.
- Assert the message names the compression: a caller shown this error should read which half of the file is wrong.

nao1215:
Three more from the same sweep, one of which names the wrong object.

An ANALYZE was read as "the first word after ANALYZE is the name", and the words each dialect writes there say how to run the statement rather than what to run it on: `ANALYZE VERBOSE t` translated to `ANALYZE VERBOSE`, so SQLite was asked to analyze something called VERBOSE and the table the caller named was dropped from the statement, and MySQL's `ANALYZE TABLE t` became `ANALYZE "TABLE"`. Those words are read and dropped now, and a statement naming more than SQLite analyzes -- several tables, a column list -- is refused by name instead of being narrowed in silence.

Five clauses standing around statements this package already carries answered with ErrInvalidSyntax, the sentinel for a query that could not be read. An index's INCLUDE columns and a view's WITH CHECK OPTION are read and dropped, which is what this package already does with CONCURRENTLY, the index method and the other clauses on those same two statements; the first changes how an index is stored rather than what a query answers, and the second constrains writes SQLite refuses through a view anyway. PostgreSQL's `SELECT ... INTO name` is the `CREATE TABLE name AS SELECT ...` that SQLite spells and now translates to it, in the TEMP spelling and in a compound query too, whose INTO PostgreSQL takes from the first select; it is read only where the query is the whole statement, so an INTO inside a subquery is still a statement that stops in the middle rather than one silently dropped. MySQL's INTO writes a file or fills a session variable and is refused by name in both the position before the FROM and the one after the query.

The third is about what a caller is told rather than what runs. A file whose suffix claims a compression its bytes do not carry answered ErrCompression with "gzip: invalid header" for gzip, xz and zlib, whose decoders read their header when the stream opens, and ErrParsing with "failed to read CSV header: magic number mismatch" for the other five, which fail on the first read -- inside the format reader, where the CSV had not been reached at all. A caller who branches on ErrCompression to tell "this file is not really compressed" from "this file is not really a CSV" got the right answer for three codecs out of eight. What a decoder reports while it is being read now carries a sentinel of its own, which the load reads as ErrCompression, so all eight answer alike; a file that decompresses and then is not the format its name claims still answers as a parse failure.

Verified with `make test`, `make lint`, the staticcheck configuration, a cross build for windows/amd64, and sqly built and tested against this checkout through a local replace. On the compressed-load benchmarks the tag costs one call per read: xz and zstd are unchanged and gzip is 1.04 times the time, with allocations unchanged.

Closes #927
Closes #928
Closes #929

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dialect translation for `ANALYZE`, including supported options and clearer handling of unsupported forms.
  * Added support for PostgreSQL `SELECT ... INTO`, translating it into table creation, including temporary and compound queries.
  * Translation now accepts and drops supported index and view clauses.
  * Compression mismatches consistently report a compression error, while invalid decompressed content remains a parsing error.
* **Documentation**
  * Documented newly supported, dropped, and rejected SQL constructs.


nao1215:
Fixed in the commit above: the query behind a WITH is the statement's own, so it now carries an INTO the way the same query without a WITH does, and `WITH c AS (SELECT 1 AS a) SELECT a INTO newt FROM c` translates to `CREATE TABLE newt AS WITH c AS (SELECT 1 AS a) SELECT a FROM c`. MySQL's INTO after such a query is refused there too. A regression case is in dialect_test.go beside the others, along with one for an INTO inside a WITH subquery, which stays a statement that stops in the middle.

PR #1033, opened 2026-09-02
docs(dialect): say that a decimal value is a floating-point value here

### LINKED ISSUE #1032: The dialect godoc does not say that a decimal value is a floating-point value here
Found on ea7b331 while asking what each type reaches in each dialect, and recorded as documentation because the behavior is right for what this package is and surprising for what a caller writes.

## Phenomenon

A value with a decimal point is a floating-point value here, and it is an exact decimal in all three source dialects, so the most ordinary arithmetic a caller can write answers something else. `SELECT 0.1 + 0.2` answers 0.30000000000000004 under mysql, postgresql and googlesql alike, where MySQL and PostgreSQL both answer 0.3; the same holds for a sum over a column of prices loaded from a CSV. Nothing in the package documentation says so. The dialect godoc has a paragraph on rounding that comes close -- it says what "a decimal literal written in the query, which MySQL reads as an exact decimal" cannot reproduce -- but it says it about ROUND's tie-breaking rather than about arithmetic, and a reader looking for "why is my total off by 4e-17" does not find it.

## Reproduction

```go
package main

import (
	"database/sql"
	"fmt"

	_ "modernc.org/sqlite"
)

func main() {
	if err := dialect.RegisterFunctions(); err != nil {
		panic(err)
	}
	defer db.Close()
	for _, name := range []string{"mysql", "postgresql", "googlesql"} {
		d, _ := dialect.Parse(name)
		out, _ := dialect.Translate(d, "SELECT 0.1 + 0.2")
		var got any
		_ = db.QueryRow(out).Scan(&got)
		fmt.Printf("%-11s 0.1 + 0.2 => %v\n", name, got)
	}
}
```

```
mysql       0.1 + 0.2 => 0.30000000000000004
postgresql  0.1 + 0.2 => 0.30000000000000004
googlesql   0.1 + 0.2 => 0.30000000000000004
```

MySQL 8.4 and PostgreSQL 17 both answer 0.3. BigQuery answers 0.30000000000000004 for two FLOAT64 literals and 0.3 for two NUMERIC ones, so the googlesql answer is right for one reading of the query and wrong for the other.

Two more observations of the same root, for whoever writes the words: `1.005::numeric(10,2)` answers 1 where PostgreSQL answers 1.01, because the float nearest 1.005 is below it and rounds down; and `1.5::numeric(10,2)` answers 1.5 where PostgreSQL answers 1.50, because a scale is a property of a decimal and SQLite has no decimal to carry it.

## Expected behavior

The behavior stays. SQLite has no decimal type, a number loaded from a file is REAL, and computing decimal arithmetic exactly would mean carrying a decimal through every operator and every helper, which is a different package. What is missing is the sentence that says so, in the dialect godoc beside the rounding paragraph it belongs with: a value with a decimal point is a binary floating-point value here, not the exact decimal MySQL and PostgreSQL hold; a sum of such values can differ from the engine's in the last bits; a scale written on a type is applied by rounding and not carried, so a cast to a scale of two answers 1.5 rather than 1.50; and a caller who needs exact decimal arithmetic should hold the value as an integer number of the smallest unit.

## Cause (guess)

Not a defect in a function: it is what SQLite's REAL is. The rounding paragraph in dialect/doc.go already names the neighbouring case, which is why the gap is worth closing there rather than in a new section.

## Tests to add

- The godoc has no test, so what holds this is the doc_sync test that already checks the documented sections exist, plus a case in dialect/internal/runtime pinning `0.1 + 0.2` under each dialect at the value above -- so that a later change to the arithmetic is a deliberate change to a documented answer rather than a silent one.
- A case pinning `CAST(1.5 AS DECIMAL(10,2))` at 1.5 in each dialect, for the same reason: the missing trailing zero is documented behavior and should fail loudly if it moves.

nao1215:
`SELECT 0.1 + 0.2` answers 0.30000000000000004 under every dialect here, where MySQL 8.4 and PostgreSQL 17 both answer 0.3. That is the arithmetic there is -- SQLite has no decimal type and a number loaded from a file is a REAL -- and it stays. What was missing is the sentence saying so, for a reader wondering why a total over a column of prices is off in the last bits.

Two more consequences of the same root are written down with it: a scale on a type is applied by rounding rather than carried, so a cast to two places answers 1.5 where both engines answer 1.50; and a value whose nearest float falls under a tie rounds with the float, so 1.005 to two places is 1 here and 1.01 in PostgreSQL. The paragraph goes beside the rounding one, which already came close -- it says what "a decimal literal written in the query, which MySQL reads as an exact decimal" cannot reproduce, but says it about ROUND's tie-breaking rather than about arithmetic.

The three answers are pinned by tests as well as written down, so a later change to any of them is a deliberate change to a documented answer rather than a silent one.

Closes #1032

## Summary by CodeRabbit

* **Documentation**
  * Documented decimal floating-point behavior, including rounding semantics and differences from exact decimal arithmetic.
  * Added guidance for preserving exact decimal values using integer-based smallest-unit representations.

* **Tests**
  * Clarified expected results for decimal rounding and scale behavior across supported dialects.
  * Documented differences between SQLite floating-point arithmetic and exact decimal results in other database systems.


nao1215:
Both suggestions taken in 1c09c00. Each pinned row now carries the engine's own answer beside it, so the want reads as this package's answer and the contrast is visible, and MySQL's tie case is pinned alongside PostgreSQL's: CAST(1.005 AS DECIMAL(10,2)) is 1 here and 1.01 in mysql:8.4, the same difference PostgreSQL shows.

PR #1043, opened 2026-09-02
test(prep): hold the two entry points to one answer

nao1215:
## Summary

`Process` and `ProcessToWriter` are two ways to ask `prep` the same question -- one hands back a reader over the processed bytes, the other writes them where the caller says -- and nothing held them equal. What would drift in silence is the count a caller reports and the rows a second pass reads.

The other half of the same invariant: what one pass writes is what a second pass reads. CSV, TSV and LTSV keep their format and JSON becomes JSONL, so the second pass is told the `OutputFormat` the first reported and has to get the same rows back. The JSON cases carry what there is to lose, since prep holds a JSON row as one raw string in a single column: a newline, a tab, a nested object, a key outside ASCII.

Both tests pass as they stand -- this is a negative result made permanent. They are here because a rule with two implementations is the one that drifts, which is where several of this package's past bugs came from.

## Verification

- `make test` green, coverage 92.7%
- `make lint` 0 issues

## Summary by CodeRabbit

* **Tests**
  * Added coverage ensuring both processing entry points produce consistent output, decoded records, errors, and metadata across supported formats and options.
  * Added round-trip coverage for reading processor output again, including JSON and JSONL data with special characters, nested values, and non-ASCII text.
* **Documentation**
  * Updated the unreleased changelog with the new test coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


nao1215:
Both findings were right and are fixed in the last commit. The agreement test now compares the error messages in order rather than the counts, so two paths reporting the same number of different errors no longer pass. The round trip names the format it expects per case, asserts it before the second pass and hands that constant to the second pass rather than the field the first pass reported, and the TSV case has two columns so output that is really CSV cannot read back as TSV.

PR #1045, opened 2026-09-02
fix(dialect): answer the clock names, and check the refusal tables against the engines

### LINKED ISSUE #1044: A bare CURRENT_TIMESTAMP answers the string "CURRENT_TIMESTAMP" under every dialect but sqlite
## What is wrong

Under `mysql`, `postgresql` and `googlesql`, a bare `CURRENT_TIMESTAMP` answers the string `"CURRENT_TIMESTAMP"` rather than the time. `CURRENT_DATE` and `CURRENT_TIME` answer their own names the same way. Nothing errors, and the column holds a plausible-looking value, so a query that stamps rows with the current time writes the same literal text into every row.

The three are the most ordinary clock expressions in SQL, and they are the standard spelling every one of these dialects prefers over `NOW()`.

## Reproduction

```go
package dialect_test

import (
	"database/sql"
	"testing"

	_ "modernc.org/sqlite"
)

func TestBareCurrentTimestampReadsTheClock(t *testing.T) {
	if err := dialect.RegisterFunctions(); err != nil {
		t.Fatal(err)
	}
	db, err := sql.Open("sqlite", ":memory:")
	if err != nil {
		t.Fatal(err)
	}
	defer db.Close()

	for _, d := range []dialect.Dialect{dialect.MySQL, dialect.PostgreSQL, dialect.GoogleSQL} {
		for _, name := range []string{"CURRENT_TIMESTAMP", "CURRENT_DATE", "CURRENT_TIME"} {
			out, err := dialect.Translate(d, "SELECT "+name)
			if err != nil {
				t.Fatalf("%s under %s: %v", name, d, err)
			}
			var got string
			if err := db.QueryRow(out).Scan(&got); err != nil {
				t.Fatalf("%s under %s: %v", name, d, err)
			}
			if got == name {
				t.Errorf("%s under %s answered its own name; translated to %q", name, d, out)
			}
		}
	}
}
```

All nine cases fail on the current main. `SELECT CURRENT_TIMESTAMP` translates to `SELECT "CURRENT_TIMESTAMP"`.

Under `--dialect sqlite`, which is the identity translation, the same query answers the time, so the fault is in the translation rather than in the engine.

## Cause

`bareValue` in `dialect/internal/lower/shared.go` turns a bare word that a dialect reads as a value into an `ast.Keyword`. Its table names two words:

```go
var bareValueNames = map[string]string{
	keywordLocalTime:      keywordCurrentTime,
	keywordLocalTimestamp: keywordCurrentTimestamp,
}
```

The comment above it says these are "the words a dialect reads as a value and SQLite does not spell the same way", and that is why the three standard ones are missing: SQLite spells them the same, so they looked like they needed no rule.

They do, because a bare word that is not in this table stays a column reference, and the renderer quotes a column reference whose name is a SQLite keyword -- which is right for a column and fatal for this. SQLite reads a double-quoted name that matches no column as a string literal rather than failing, so `"CURRENT_TIMESTAMP"` becomes the eleven characters of its own name and nothing reports it.

That last step is also why the neighbouring cases are visible rather than silent: `CURRENT_USER`, `SESSION_USER` and the rest are not SQLite keywords, so they are rendered unquoted and fail with "no such column: CURRENT_USER". Only the three SQLite has a keyword for are quoted, and only they are silent.

## Expected behavior

`CURRENT_DATE`, `CURRENT_TIME` and `CURRENT_TIMESTAMP` written as bare words read the clock, under every dialect, as they already do written as `CURRENT_TIMESTAMP()` and as `LOCALTIMESTAMP` does today.

A column that is really named one of these is still reachable: `bareValue` leaves a quoted name alone, and all three are reserved words in MySQL and PostgreSQL, so a caller with such a column quotes it there too.

## How to keep it fixed

A test over the three names and every dialect, asserting the answer is not the name itself -- that is what makes the silence visible, since any correct answer is a timestamp and no correct answer is the word.

Worth pairing with the neighbouring case, which is a worse message rather than a wrong answer: a bare `CURRENT_USER` under any dialect reaches SQLite as a column reference and fails with "no such column", where this package refuses `CURRENT_USER()` by name and says why. The two spellings should answer the same.

nao1215:
## Summary

This is the first of the two things we agreed to do: stop building the tables of refused function names by sweeping an engine by hand, and make their completeness something CI can check.

A sweep is only as complete as the person running it. The one behind these tables walked each engine's scalar functions and never reached the aggregates, which is how `PERCENTILE_CONT`, `MAX_BY` and three others went on reaching SQLite until a second sweep found them last week. Asking the engine for its own catalog is the same question with nobody to forget half of it.

`dialect/testdata/engine_functions_mysql.txt` and `engine_functions_postgresql.txt` now hold what MySQL's help tables and PostgreSQL's `pg_proc` say each engine defines -- 391 and 808 names -- written by `scripts/dump-engine-functions.sh`, which starts the two containers and runs the queries. `TestEveryEngineFunctionHasAnAnswer` requires that a query naming any of them is translated into something SQLite can run, or refused here by name, and never handed to SQLite to fail as "no such function: NAME", which tells a caller that a name they did write does not exist.

A name a later engine adds appears when those files are regenerated, and the test fails on it then. BigQuery publishes no catalog of its functions, so GoogleSQL's table stays hand-written and nothing checks it this way; that is said in the package documentation rather than left to be discovered.

## What the check found

150 MySQL names and 571 PostgreSQL names reached SQLite. Most are families -- spatial, the data dictionary, the performance schema, cryptography, catalog and administration, sequences, large objects, XML, text search, ranges, network addresses, geometry, the array and record functions, and the names each engine keeps for its own arithmetic -- and are now refused with the reason their family gives. A family too large to list is written as a prefix, and a prefix never claims a name this package registers a helper for, which is what keeps `PG_TYPEOF` working while everything else beginning with `PG_` is refused.

Five were real gaps, and one of them is the reason this PR is not just tables:

**A bare `CURRENT_TIMESTAMP` answered the string `"CURRENT_TIMESTAMP"`** under every dialect but `sqlite`, and `CURRENT_DATE` and `CURRENT_TIME` answered theirs, with nothing to report it. A query stamping rows with the current time wrote the same eleven characters into every row. The three are the standard spelling each of these dialects prefers over `NOW()`. Filed as #1044 with the mechanism: a bare word not in the table of words that stand for a value stays a column reference, a column reference whose name is a SQLite keyword is rendered quoted, and SQLite reads a quoted name matching no column as a string rather than refusing it -- which is also why the neighbouring `CURRENT_USER` fails loudly instead.

The other four: MySQL's `UTC_TIMESTAMP`, `UTC_DATE`, `UTC_TIME` and `SYSDATE` had no answer at all; `BINARY(x)` was not lowered where `BINARY x` was; a clock keyword given the fractional-seconds precision MySQL and PostgreSQL allow reached SQLite as a name it does not have; and `POSITION` or `CONVERT` with the wrong number of arguments did the same.

## What is left

MySQL has no name without an answer. PostgreSQL has 24, listed in `dialect/testdata/untranslated_postgresql.txt` with what they are -- `DATE_ADD`, `REGEXP_MATCH`, `TO_BIN`, `OVERLAPS` and the rest, each checked against a running PostgreSQL 17. The test fails when a name not on that list joins them and when a name on it gains an answer, so the debt is visible and can only shrink. Working through it is the follow-up.

There is a structural alternative worth considering separately: rather than tables of what to refuse, refuse anything whose translated name is neither a SQLite builtin nor a helper this package registers. That closes the class for all three dialects at once, including names no catalog lists, and it would need SQLite's own function list generated into the lowering layer. It would also refuse a function a caller registered on their own connection, which is why it is a decision rather than a cleanup.

## Verification

- `make test` green, coverage 92.7%; `make lint` 0 issues
- The ratchet was checked in both directions: taking a name out of the untranslated list fails, and putting a name that works into it fails
- No name that worked before stopped working: the classification of all 1,199 names was captured before and after each change and diffed
- `UTC_TIMESTAMP`, `UTC_DATE`, `UTC_TIME`, `SYSDATE`, `NOW` and `BINARY(1)` were compared against a running MySQL 8.4, and the 24 remaining PostgreSQL names against a running PostgreSQL 17

Closes #1044

## Summary by CodeRabbit

- **Bug Fixes**
  - Bare `CURRENT_TIMESTAMP`, `CURRENT_DATE`, and `CURRENT_TIME` now correctly return clock values across supported dialects.
  - Added support for MySQL `UTC_TIMESTAMP`, `UTC_DATE`, `UTC_TIME`, and `SYSDATE`.
  - MySQL `BINARY(x)` now works alongside the existing `BINARY x` syntax.
  - Invalid argument counts for `BINARY`, `POSITION`, and `CONVERT` are now rejected with clear errors.
  - Unsupported clock precision is rejected explicitly.
  - Additional MySQL and PostgreSQL functions are now identified as unsupported instead of reaching SQLite and failing with unknown-function errors.


nao1215:
Right, and thank you -- the trap would have removed a container this script did not create, which is somebody else's local state. The names now carry the process id, so the trap can only ever remove what this run started.

Two things came with it: the wait loop used to fall through when the engines never came up, leaving the queries to fail under their own errors, so it now says so and writes nothing. And the script was run end to end against the two containers: it reproduces the committed corpora byte for byte, which is the check that the files in testdata are what the script produces rather than what a session happened to leave behind.

nao1215:
Right again, and this time at the root: no name this script picks is safe, because any name can already be taken, and a process id is reused. The containers are now addressed by the ids `docker run` returns, so the trap can only remove what this run started -- and removes nothing when a run fails before starting them. The polling flag is initialized before the loop.

The script was run end to end again afterwards: it reproduces the committed corpora byte for byte.

PR #1050, opened 2026-09-02
fix(reader): read a workbook the way its writer wrote it

### LINKED ISSUE #1047: A number whose format draws nothing is lost when it sits in the last column of a sheet
Found at v0.56.0 (60ff1b6) by loading workbooks whose cells are formatted in ways Excel users use to hide a helper column, and comparing what loads against what the file stores.

## Phenomenon

A numeric cell wearing a number format that draws nothing -- `;;;`, the format Excel users pick to hide a cell's content while keeping its value -- loads as the number it stores when it sits in a middle column and loads as the empty string when it sits in the last column of the sheet, or in any run of such cells at the end of a row. Nothing errors. A workbook with a hidden `secret` column at its right edge loads that column as an all-empty TEXT column, so the values are gone and the column's type is wrong, while the same workbook with the hidden column moved one place left loads it as REAL with every value intact. The rule the loader documents and applies in the middle of a row, that a number loads as the number the file stores whatever its format draws, is broken exactly at the row's end, which is where a helper column usually is.

## Reproduction

```go
package main

import (
	"context"
	"database/sql"
	"fmt"
	"os"
	"path/filepath"

	"github.com/nao1215/filesql"
	"github.com/xuri/excelize/v2"
)

func write(path string, header []any, records [][]any, hiddenColumn int) {
	f := excelize.NewFile()
	defer f.Close()
	f.SetSheetName("Sheet1", "data")
	hidden := ";;;"
	style, _ := f.NewStyle(&excelize.Style{CustomNumFmt: &hidden})
	f.SetSheetRow("data", "A1", &header)
	for i, row := range records {
		cell, _ := excelize.CoordinatesToCellName(1, i+2)
		f.SetSheetRow("data", cell, &row)
		axis, _ := excelize.CoordinatesToCellName(hiddenColumn, i+2)
		f.SetCellStyle("data", axis, axis, style)
	}
	f.SaveAs(path)
}

func show(path string) {
	db, _ := filesql.Open(context.Background(), path)
	defer db.Close()
	rows, _ := db.Query(`SELECT * FROM data`)
	defer rows.Close()
	cols, _ := rows.Columns()
	for rows.Next() {
		vals := make([]sql.NullString, len(cols))
		ptrs := make([]any, len(cols))
		for i := range vals {
			ptrs[i] = &vals[i]
		}
		rows.Scan(ptrs...)
		fmt.Println(cols, vals)
	}
}

func main() {
	dir, _ := os.MkdirTemp("", "hidden")
	defer os.RemoveAll(dir)

	last := filepath.Join(dir, "last", "data.xlsx")
	os.MkdirAll(filepath.Dir(last), 0o750)
	write(last, []any{"id", "name", "secret"}, [][]any{{1, "alice", 99}, {2, "bob", 98}}, 3)
	show(last)

	middle := filepath.Join(dir, "middle", "data.xlsx")
	os.MkdirAll(filepath.Dir(middle), 0o750)
	write(middle, []any{"id", "secret", "name"}, [][]any{{1, 99, "alice"}, {2, 98, "bob"}}, 2)
	show(middle)
}
```

```
[id name secret] [{1 true} {alice true} { true}]
[id name secret] [{2 true} {bob true} { true}]
[id secret name] [{1 true} {99 true} {alice true}]
[id secret name] [{2 true} {98 true} {bob true}]
```

The `secret` column holds `""` in the first workbook and `99` and `98` in the second; the two workbooks store the same numbers under the same format and differ only in which column the format is on. `PRAGMA table_info` reports the column as TEXT in the first and REAL in the second.

## Expected behavior

Both workbooks load `secret` as `99` and `98`, a REAL column. `doc.go` under `# Column Types` and the reader's own comments state the rule: a number format says how a spreadsheet paints a number and not what the number is, so a numeric cell loads as the number the file stores, with only a date, a time of day and an elapsed duration keeping their drawing. A hidden format is a number format like any other, and the middle-column case already follows the rule. A cell that stores no value at all is a different thing and should still load as empty.

## Cause (guess)

`internal/reader/xlsx.go` `ReadSheet` asks the library for the rendered rows and then `NormalizeCells` overwrites, in place, the cells the sheet's own XML says store a number. The library's row reader (`excelize` `rowXMLHandler`) appends a cell to a row only when its rendered value is non-empty or it holds a formula, so a cell whose format renders `""` is dropped, and when nothing follows it in the row the row simply ends one cell short; a dropped cell in the middle is padded back in by the cells after it. `NormalizeCells` then skips any XML cell whose column is at or past `len(rows[r])` (`if r < 0 || r >= len(rows) || c < 0 || c >= len(rows[r]) { continue }`), so a hidden number in the last column has nowhere to land, while one in the middle lands on the padding. A row whose every cell is hidden is the extreme case: the library returns no cells for it, the row is later recognized as held (`heldRows`) and becomes a record of empties, and its numbers are lost too. The library-side fallback `normalizeXLSXNumbers` (`xlsx_number.go`) has the same shape of gap, since it walks only the cells the library returned and skips `""`. The fix is for `NormalizeCells` to extend a row (and the row list) out to the cells the XML names rather than skipping them, and for the fallback to read the raw row shape rather than the rendered one when deciding which cells exist.

## Tests to add

- In `internal/reader/xlsx_test.go`, a workbook written with excelize whose last column wears `;;;` on numbers: the column loads as the numbers, with type REAL or INTEGER, and the same workbook with the column moved to the middle loads identically apart from column order.
- The sibling cases: a row whose every cell wears `;;;` loads as its numbers rather than as a record of empties, and a header row is unaffected (a hidden header is text and keeps drawing nothing, which is the existing rule for strings).
- In `internal/reader/xlsx_date_stream_test.go`, the equality test between the XML reading and the library reading (`NormalizeCells` against `normalizeXLSXNumbers`) gains a sheet with a trailing hidden cell, so the two readings cannot drift on this again.
- In `autosave_test.go`, an in-place save of such a workbook with no edits leaves the file's cells unchanged, and one with an edit to a visible cell leaves the hidden column's values and format alone, since `xlsxSheetBefore` reads the sheet through the same normalization and a fix there changes what a save compares against.

### LINKED ISSUE #1048: A sheet spelled with a namespace prefix, single-quoted attributes or a line break before r= drops records of empty cells or invents them
Found at v0.56.0 (60ff1b6) by rewriting the worksheet XML of a workbook excelize wrote into spellings other writers use and comparing what loads.

## Phenomenon

The reading that decides which rows of a sheet hold a cell is a byte scan of the worksheet XML that recognizes only the spelling Excel and excelize use: an unprefixed `<row` and `<c`, and an `r` attribute written as a space followed by `r="`. A worksheet whose elements carry a namespace prefix, as the Open XML SDK writes them (`<x:row r="3"><x:c r="A3" t="s">`), reports no row as holding a cell, so every record whose cells all hold the empty string is silently dropped from the table. A worksheet whose attributes are single-quoted (`<row r='3'>`, which is valid XML and which Excel and excelize both open) or whose `r` attribute is preceded by a line break or a tab rather than a space, as a pretty-printing writer emits, is read with sequential row numbers instead of the ones written, so a row the sheet does not hold becomes a record of empties and the record that really sits at the end of the sheet is lost. In every case `Open` succeeds and the table has a plausible shape, so a caller has nothing to notice. The library the loader is built on reads all three spellings correctly, and so does Excel; only this scan does not.

## Reproduction

```go
package main

import (
	"archive/zip"
	"bytes"
	"context"
	"database/sql"
	"fmt"
	"io"
	"os"
	"path/filepath"
	"regexp"
	"strings"

	"github.com/nao1215/filesql"
	"github.com/xuri/excelize/v2"
)

// build writes a sheet where row 3 and row 7 hold the empty string in every
// cell (records), and row 5 is absent from the sheet (not a record).
func build(dir, name string) string {
	path := filepath.Join(dir, name, "data.xlsx")
	os.MkdirAll(filepath.Dir(path), 0o750)
	f := excelize.NewFile()
	defer f.Close()
	f.SetSheetName("Sheet1", "data")
	f.SetSheetRow("data", "A1", &[]any{"id", "name"})
	f.SetSheetRow("data", "A2", &[]any{1, "alice"})
	f.SetSheetRow("data", "A3", &[]any{"", ""})
	f.SetSheetRow("data", "A4", &[]any{2, "bob"})
	f.SetSheetRow("data", "A6", &[]any{3, "carol"})
	f.SetSheetRow("data", "A7", &[]any{"", ""})
	f.SaveAs(path)
	return path
}

func rewriteSheet(path string, fn func(string) string) {
	data, _ := os.ReadFile(path)
	zr, _ := zip.NewReader(bytes.NewReader(data), int64(len(data)))
	var buf bytes.Buffer
	zw := zip.NewWriter(&buf)
	for _, file := range zr.File {
		rc, _ := file.Open()
		body, _ := io.ReadAll(rc)
		rc.Close()
		if file.Name == "xl/worksheets/sheet1.xml" {
			head, rest, _ := strings.Cut(string(body), "?>")
			body = []byte(head + "?>" + fn(rest))
		}
		w, _ := zw.Create(file.Name)
		w.Write(body)
	}
	zw.Close()
	os.WriteFile(path, buf.Bytes(), 0o600)
}

func show(label, path string) {
	db, err := filesql.Open(context.Background(), path)
	if err != nil {
		fmt.Println(label, err)
		return
	}
	defer db.Close()
	rows, _ := db.Query(`SELECT * FROM data`)
	defer rows.Close()
	var out []string
	for rows.Next() {
		var id sql.NullString
		var name string
		rows.Scan(&id, &name)
		out = append(out, fmt.Sprintf("[%s %q]", id.String, name))
	}
	fmt.Println(label, strings.Join(out, " "))
}

func main() {
	dir, _ := os.MkdirTemp("", "spellings")
	defer os.RemoveAll(dir)

	show("as excelize wrote it:      ", build(dir, "base"))

	prefixed := build(dir, "prefixed")
	element := regexp.MustCompile(`<(/?)([A-Za-z])`)
	rewriteSheet(prefixed, func(s string) string {
		s = element.ReplaceAllString(s, "<${1}x:${2}")
		return strings.Replace(s, `xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"`, `xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main"`, 1)
	})
	show("elements prefixed x:       ", prefixed)

	single := build(dir, "single")
	attribute := regexp.MustCompile(`="([^"]*)"`)
	rewriteSheet(single, func(s string) string { return attribute.ReplaceAllString(s, "='${1}'") })
	show("attributes single-quoted:  ", single)

	tabbed := build(dir, "tabbed")
	rewriteSheet(tabbed, func(s string) string { return strings.ReplaceAll(s, ` r="`, "\n\tr=\"") })
	show("r attribute on its own line:", tabbed)
}
```

```
as excelize wrote it:       [1 "alice"] [ ""] [2 "bob"] [3 "carol"] [ ""]
elements prefixed x:        [1 "alice"] [2 "bob"] [3 "carol"]
attributes single-quoted:   [1 "alice"] [ ""] [2 "bob"] [ ""] [3 "carol"]
r attribute on its own line: [1 "alice"] [ ""] [2 "bob"] [ ""] [3 "carol"]
```

The first line is the table the sheet describes: five records, two of them empty, with the absent row 5 contributing nothing. The prefixed spelling drops both empty records. The single-quoted and line-broken spellings put an empty record where row 5 is absent and drop the empty record at row 7, so the table has the right count with the wrong rows in it. Rewriting the sheet with `xml.Unmarshal` or opening it in Excel gives the first table for all four files.

## Expected behavior

All four files load the same five records. The rule is stated in `internal/reader/xlsx.go` above `tableRows`: a row holding no cell at all is not a record, and a row whose cells are present and empty is one, which the file says and the file is asked. Which spelling of XML the file uses is not part of that rule, and the library that opens the workbook, being an XML parser, is indifferent to it; a byte scan that stands in for the parser has to accept what the parser accepts, at least for the constructs it looks at: an optional namespace prefix on the element name, either quote character, and any XML whitespace between attributes and around the `=`.

## Cause (guess)

`internal/reader/xlsx_date_stream.go` `scanRowWindow` matches tags with `bytes.HasPrefix(tag, "<row")` and `bytes.HasPrefix(tag, "<c")`, which a prefixed `<x:row` never satisfies, and `rowNumberAttr` finds the attribute with `bytes.Index(tag, ` r="`)`, which single quotes, a tab or a line break before `r`, and whitespace around `=` all defeat; a row whose number is not found is counted as the row after the previous one, which is right only for a sheet with no gaps. The scan is deliberately not an XML parser, for the cost reason its comment gives, and that is fine; it needs to skip an optional `name:` prefix after `<` before comparing the local name, and to find `r` as an attribute name delimited by XML whitespace, followed by optional whitespace, `=`, optional whitespace and either quote. The sibling reading in the same file, `scanSheetValues`, goes through `encoding/xml` and is unaffected, which is why the values of the rows that survive are right and only the row set is wrong.

## Tests to add

- In `internal/reader/xlsx_date_stream_test.go`, `scanSheetRows` over the same sheet in four spellings -- as excelize writes it, with every element prefixed `x:`, with single-quoted attributes, with each `r` attribute on its own line behind a tab and with spaces around `=` -- answers the same row set, including a gap.
- In `internal/reader/xlsx_test.go`, a workbook rewritten into each spelling loads the same records as the original through `ReadSheet`, with a row of empty strings in the middle, a row absent from the sheet, and a row of empty strings at the end, which are the three cases the scan's answer changes.
- A test that the two scans over one sheet agree: for each spelling, the rows `scanSheetRows` reports and the rows `scanSheetValues` visits (which reads the same `r` attributes through `encoding/xml`) are the same set, so the byte scan cannot again accept less than the parser does without a test saying so.

### LINKED ISSUE #1049: A workbook whose main part is not at xl/workbook.xml is read by the fallback, which drops hidden numbers and records of empty cells
Found at v0.56.0 (60ff1b6) by moving the workbook part of a workbook excelize wrote to a path other than `xl/workbook.xml`, which the package format allows and the library opens, and comparing what loads.

## Phenomenon

The XLSX loader has two readings of a sheet: the one it uses, which walks the sheet's own XML, and a fallback that asks the library about each cell, taken when the workbook's bytes do not say where the sheet is. The fallback is reached for any workbook whose main part is not at `xl/workbook.xml`, because the part is looked up at that fixed path rather than through the package's root relationships, and the fallback does not answer the same as the reading it stands in for: a numeric cell whose format draws nothing (`;;;`) loads as NULL in every column under the fallback where the primary reading loads the number, and a record whose cells all hold the empty string is dropped under the fallback where the primary reading keeps it. Nothing says which reading a load took, so the same rows in the same sheet load differently depending on where inside the zip the writer put the workbook part, and a caller comparing two such files sees rows and values that are not there.

## Reproduction

```go
package main

import (
	"archive/zip"
	"bytes"
	"context"
	"database/sql"
	"fmt"
	"io"
	"os"
	"path/filepath"

	"github.com/nao1215/filesql"
	"github.com/xuri/excelize/v2"
)

func show(label, path string) {
	db, err := filesql.Open(context.Background(), path)
	if err != nil {
		fmt.Println(label, err)
		return
	}
	defer db.Close()
	rows, _ := db.Query(`SELECT * FROM data`)
	defer rows.Close()
	for rows.Next() {
		var id, secret sql.NullString
		var name string
		rows.Scan(&id, &secret, &name)
		fmt.Printf("%s [%s %s %q]\n", label, id.String, secret.String, name)
	}
}

func main() {
	dir, _ := os.MkdirTemp("", "part")
	defer os.RemoveAll(dir)
	path := filepath.Join(dir, "data.xlsx")

	f := excelize.NewFile()
	f.SetSheetName("Sheet1", "data")
	hidden := ";;;"
	style, _ := f.NewStyle(&excelize.Style{CustomNumFmt: &hidden})
	f.SetSheetRow("data", "A1", &[]any{"id", "secret", "name"})
	f.SetSheetRow("data", "A2", &[]any{1, 99, "alice"})
	f.SetSheetRow("data", "A3", &[]any{"", "", ""})
	f.SetSheetRow("data", "A4", &[]any{2, 98, "bob"})
	f.SetCellStyle("data", "B2", "B4", style)
	f.SaveAs(path)
	f.Close()
	show("workbook at xl/workbook.xml:", path)

	data, _ := os.ReadFile(path)
	zr, _ := zip.NewReader(bytes.NewReader(data), int64(len(data)))
	var buf bytes.Buffer
	zw := zip.NewWriter(&buf)
	for _, file := range zr.File {
		rc, _ := file.Open()
		body, _ := io.ReadAll(rc)
		rc.Close()
		name := file.Name
		switch name {
		case "xl/workbook.xml":
			name = "xl/book.xml"
		case "xl/_rels/workbook.xml.rels":
			name = "xl/_rels/book.xml.rels"
		case "[Content_Types].xml", "_rels/.rels":
			body = bytes.ReplaceAll(body, []byte("xl/workbook.xml"), []byte("xl/book.xml"))
		}
		w, _ := zw.Create(name)
		w.Write(body)
	}
	zw.Close()
	os.WriteFile(path, buf.Bytes(), 0o600)
	show("workbook at xl/book.xml:    ", path)
}
```

```
workbook at xl/workbook.xml: [1 99 "alice"]
workbook at xl/workbook.xml: [  ""]
workbook at xl/workbook.xml: [2 98 "bob"]
workbook at xl/book.xml:     [1  "alice"]
workbook at xl/book.xml:     [2  "bob"]
```

The second file is the first with its workbook part renamed and the two references to it updated, which is a change the library reads through without complaint; it loads with the hidden numbers gone and the empty record gone.

## Expected behavior

Both files load three records with `secret` holding 99, NULL and 98. Where the workbook part sits is bookkeeping the package's `_rels/.rels` answers, and a loader that reads it should ask; and whichever reading a load takes, a number whose format draws nothing loads as the number, per the rule in `doc.go` `# Column Types` and in the reader, and a row of empty cells is a record, per `tableRows`. The fallback exists so that a sheet is not read with cells missing when the fast reading cannot find it; it should not read the sheet with rows and values missing instead.

## Cause (guess)

Two things, in `internal/reader/xlsx_date_stream.go` and `xlsx_number.go`. `sheetPart` reads `xl/workbook.xml` and `xl/_rels/workbook.xml.rels` by literal name; when either is absent it reports the sheet as not found, and both `cellValuesFromXML` and `rowsHoldingCellsFromXML` report false, so `NormalizeCells` falls back to `normalizeXLSXDates(normalizeXLSXNumbers(...))` and `heldRows.set` returns nil. The fix for that half is to find the workbook part through the `officeDocument` relationship in `_rels/.rels`, read its rels at `<dir>/_rels/<base>.rels`, and resolve a sheet target relative to the workbook part's directory, which is what the library does. The other half is that the fallback disagrees with the primary reading: `normalizeXLSXNumbers` walks only the cells the library rendered and skips any rendered as `""`, so a hidden number is never rewritten, and a nil `heldRows` set answers false for every row, so a row of empty cells is treated as absent. Finding the sheet through the rels makes the fallback rare, but the test that holds the two readings equal should cover these two inputs so that whatever still reaches the fallback (a style table too long to walk is the remaining trigger) loads the same.

## Tests to add

- In `internal/reader/xlsx_date_stream_test.go`, `sheetPart` finds the sheet of a workbook whose main part is at `xl/book.xml`, at `book/wb.xml` with its sheets under `book/sheets/`, and at the standard path, and resolves a target written relative to the workbook part, written from the root with a leading `/`, and written with a leading `./`.
- In `internal/reader/xlsx_test.go`, the workbook above loads the same three records before and after its workbook part is renamed.
- The existing test that holds `NormalizeCells` and the library-side reading to one answer gains a sheet with a `;;;` number in a middle column and a row of empty strings, so the fallback cannot answer differently for them.

nao1215:
## Summary

Three fixes to the XLSX loader, found by writing the same sheet the ways a writer other than Excel writes it and comparing what loads. Each is a place where the loader's own reading of a workbook -- the sheet's XML, which the loader reads beside the library for the cells whose drawing is not their value -- accepted less than the library and Excel do, or reached a cell the library had dropped and had nowhere to put it.

- A numeric cell wearing `;;;`, the format that hides a column while keeping its values, loaded as the number in a middle column and as the empty string in the last column, or in a run of such cells at the end of a row. The library returns no cell for one, and only the cells after a dropped cell padded it back into the row. The rewrite from the XML now extends a row out to the cells the sheet holds, and the rows out to a row made entirely of such cells; the library-side fallback walks the stored cells so it answers the same; the in-place save takes its extent from the normalized rows.
- The byte scan that says which rows hold a cell recognized only Excel's spelling of the XML. Namespace-prefixed elements, as the Open XML SDK writes them, dropped every record of empty cells; single-quoted attributes or a line break before `r=` numbered the rows in sequence, which invented a record where the sheet holds no row and lost the one at its end. The scan reads a tag's name whole, drops a namespace prefix, and finds an attribute with any XML whitespace and either quote.
- The sheet's XML was looked for through `xl/workbook.xml` by name, so a workbook whose main part sits elsewhere fell back to the cell-by-cell reading, which dropped records of empty cells and hidden numbers. The lookup follows `_rels/.rels` to the main part and the main part's own rels to each sheet, the way the library does.

Closes #1047, closes #1048, closes #1049.

## Tests

- `internal/reader/xlsx_test.go`: a hidden column in the last, a middle and every column; the same sheet in four spellings; the main part at three locations.
- `internal/reader/xlsx_date_stream_test.go`: the byte scan over each spelling, the byte scan and the XML scan held to the same rows under each, and `sheetPart` over seven package layouts.
- `internal/reader/xlsx_date_test.go`: the two cell readings held to one answer now cover a hidden cell with a visible one after it, at the end of a row, and a row of nothing else.
- `autosave_test.go`: an in-place save that edits nothing leaves a hidden column's cells and format alone, and one that edits a visible cell writes only that cell.

## Performance

`BenchmarkOpenFormattedWorkbook`, `BenchmarkOverwriteWorkbookOfNumbers` and `BenchmarkOverwriteWorkbook` at `-count=5` before and after, medians: 170 ms → 163 ms, 135 ms → 134 ms, 238 ms → 222 ms; B/op unchanged, allocs/op +200 of 2.7 M. The byte scan runs only when a sheet is asked which rows hold a cell, and the row extension touches only the cells the XML names.

## Compatibility

sqly builds and its tests pass against this branch with a local `replace`. No exported name changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed loading of numeric cells with formats that render no visible content.
  * Improved compatibility with spreadsheets generated by non-Excel tools, including varied XML formatting and namespaces.
  * Fixed reading workbooks whose main files are stored at nonstandard package locations.
  * Improved preservation of hidden cells, row boundaries, empty records, and worksheet formatting during XLSX saves.
  * Prevented malformed or oversized XML tags from disrupting worksheet loading.


nao1215:
Two review notes on `internal/reader/xlsx_date_stream.go`, both outside the diff:

- A comment quoting row markup: fixed in 3fcbbd6. The scan now carries whether it is inside a comment across windows and skips to `-->`, and the regression cases cover a comment between two rows and one that spans two windows with a quoted row in its middle.
- Rejecting a tag with a malformed attribute after `r`: left as is. The library parses the same sheet through `encoding/xml` before this scan is ever asked, so a sheet with a malformed attribute fails `GetRows` and never reaches it; making the scan a validator would duplicate the parser, and its comment says it reads attributes as the parser accepts them rather than that it refuses what the parser refuses.

PR #1070, opened 2026-09-03
docs: hold the documentation to what the code does

### LINKED ISSUE #1051: The nightly fuzz workflow names a package that moved, so two targets have not run since v0.55.0
Found at v0.56.0 (73d9e60) from the nightly run of 2026-09-02, https://github.com/nao1215/filesql/actions/runs/33671513163.

## Phenomenon

The Fuzz workflow names `./parser` for two of its seven targets, and that package moved to `./internal/parser` in v0.55.0. Both jobs have failed with `stat .../parser: directory not found` on every nightly run since, so `FuzzParseFormats` and `FuzzParseBinary` have not fuzzed anything for the releases in between: the two targets that read a file of untrusted bytes, one for the text formats and one for Parquet. Nothing said so, because `fail-fast: false` lets the other five pass and a scheduled run's failure notifies nobody who is watching a pull request. The workflow is the only thing that runs these targets past their seed corpus, so what looked like coverage was the seeds and no search.

## Reproduction

```bash
cd $(mktemp -d) && git clone --depth 1 https://github.com/nao1215/filesql && cd filesql
go test ./parser -run '^$' -fuzz '^FuzzParseFormats$' -fuzztime 10s
```

```
# ./parser
stat /tmp/.../filesql/parser: directory not found
FAIL	./parser [setup failed]
```

```bash
grep -rn "^func Fuzz" --include=*_test.go . | sed 's/:func /  /'
```

```
dialect/fuzz_test.go:17  FuzzTranslate(f *testing.F) {
dialect/internal/parser/parser_fuzz_test.go:16  FuzzParse(f *testing.F) {
dialect/internal/parser/parser_fuzz_test.go:110  FuzzLex(f *testing.F) {
internal/codec/codec_fuzz_test.go:65  FuzzCodecReader(f *testing.F) {
internal/parser/parser_test.go:793  FuzzParseFormats(f *testing.F) {
internal/parser/parquet_test.go:207  FuzzParseBinary(f *testing.F) {
internal/reader/csv_test.go:179  FuzzCSVReader(f *testing.F) {
```

The workflow's matrix says `./parser` for the two targets the grep places in `./internal/parser`.

## Expected behavior

Both jobs run for their ten minutes against `./internal/parser`. More generally, a job that names a package path that does not exist should not be able to sit in the matrix unnoticed: every fuzz target in the module should be in the matrix, and every matrix entry should name a package that holds the target it names.

## Cause (guess)

The package move in v0.55.0 (`refactor(api)!: stop publishing the parser package`, 6aa8931) updated the imports the compiler checks and did not update `.github/workflows/fuzz.yml`, which nothing compiles. The same class of drift can hit any workflow that names a path in a `run:` line.

## Tests to add

- A Go test that reads `.github/workflows/fuzz.yml`, collects the `package`/`target` pairs of the matrix, and compares them against the fuzz targets it finds by walking the module's `_test.go` files: it fails when a target has no matrix entry, when an entry names a target the package does not hold, and when an entry names a package directory that does not exist. `release_metadata_test.go` is the file for it, since it already holds the tests that read repository metadata rather than package behavior.
- The same test should cover the other workflows' package paths, since the fault is "a workflow names a path the compiler never checks" rather than anything about fuzzing.

### LINKED ISSUE #1052: Six code snippets in the documentation call Open and DumpDatabase without the context they take
Found at v0.56.0 (73d9e60) by walking every code block inside a doc comment and checking each call against the signature the package exports.

## Phenomenon

Six code snippets in the package documentation call `Open` or `DumpDatabase` without the `context.Context` those functions have taken since v0.55.0. They are what a reader copies out of pkg.go.dev, and none of them compiles: two omit the argument entirely and four pass `db` or a path where the context goes, which is a type error rather than a missing one. The snippets sit on `DumpDatabase` itself, on `DumpOptions`, and in the ACH section, so a caller reading the documentation of the function they are about to call is handed code that does not build.

## Reproduction

```bash
grep -rn 'DumpDatabase(db\|filesql.Open("' --include=*.go /path/to/filesql | grep -v _test.go
```

```
ach.go:59://	filesql.DumpDatabase(db, "./output") // ACH tables → .ach files, others → CSV/etc
ach.go:101://	db, _ := filesql.Open("payment.ach")
dump_options.go:124://	err := DumpDatabase(db, "./output", options)
filesql.go:82://	err := filesql.DumpDatabase(db, "./output")
filesql.go:89://	err := DumpDatabase(db, "./output")
filesql.go:95://	err := DumpDatabase(db, "./output", options)
```

The current signatures are `func Open(ctx context.Context, paths ...string) (*sql.DB, error)` and `func DumpDatabase(ctx context.Context, db *sql.DB, outputDir string, opts ...DumpOptions) error`, so each of the six is either short an argument or passing one of the wrong type.

## Expected behavior

Every snippet compiles: `filesql.Open(ctx, "payment.ach")` and `filesql.DumpDatabase(ctx, db, "./output")`, with the options argument last where one is shown. A reader should be able to paste any of them beside a `ctx` and build.

## Cause (guess)

v0.55.0 removed `OpenContext` and `DumpDatabaseContext` and moved the context onto `Open` and `DumpDatabase`. The call sites in code were updated because the compiler required it; the ones inside comments were not, because nothing checks them.

## Tests to add

- A test that finds every code block in every doc comment in the module, extracts the calls to this module's own exported functions, and fails when a call passes fewer arguments than the function takes or passes something that is not a context where the first parameter is a `context.Context`. `doc_sync_test.go` is the file for it; it already holds the tests that compare the documentation against the code. The check is worth having beyond this one fix, because the next signature change will leave the same kind of debris.
- The sibling case the same test should catch: a snippet naming a symbol the package no longer exports at all.

### LINKED ISSUE #1053: NewDumpOptions offers a compression a save refuses, and the format lists leave out XLSX
Found at v0.56.0 (73d9e60) by calling every value the option lists name and comparing what a dump does with what the list says.

## Phenomenon

`NewDumpOptions`'s godoc tells a caller to modify the options with "WithCompression(): Add compression (GZ, BZ2, XZ, ZSTD)". A save asked for BZ2 fails: `WithCompression`'s own godoc says "read only, writing not supported", and `CompressionType.NewWriter` refuses it with `ErrUnsupportedFormat`. So the summary offers one of four compressions that cannot be used, and the reader who follows it meets a failed export rather than a compressed file. The same summary says the formats are "(CSV, TSV, LTSV)", and `WithFormat`'s own list adds Parquet but not XLSX, although a dump writes a workbook when asked for `OutputFormatXLSX`. Two of the five formats a caller can actually ask for appear in no list a reader of these two functions sees.

## Reproduction

```go
db, _ := filesql.Open(ctx, "data.csv")
err := filesql.DumpDatabase(ctx, db, dir, filesql.NewDumpOptions().WithCompression(filesql.CompressionBZ2))
```

```
filesql: I/O operation failed: failed to export table d: filesql: unsupported file format: bzip2 compression is not supported for writing
```

```go
err := filesql.DumpDatabase(ctx, db, dir, filesql.NewDumpOptions().WithFormat(filesql.OutputFormatXLSX))
```

```
err=<nil> files=[d.xlsx]
```

The full sweep over the enumerations, run against the same one-table database: CSV, TSV, LTSV, Parquet and XLSX all write a file; ACH and FedWire are refused with "unsupported output format"; every compression writes except BZ2, which is refused.

## Expected behavior

`NewDumpOptions` either names what a save accepts or does not enumerate at all. BZ2 does not belong in a list of compressions to modify the options with, since a save refuses it; XLSX belongs in the format list, since a save writes it. The two lists in `WithFormat` and `WithCompression` are the detailed ones and should be the only place the vocabulary is spelled out, so the summary above them cannot drift from them again.

## Cause (guess)

The two lists were written when the package wrote three formats and four compressions, and each later format and codec was added to the enumeration and to `WithFormat`/`WithCompression` without the summary being revisited. The BZ2 entry is older than the discovery that the library this package uses has no bzip2 writer; `WithCompression` was corrected then and `NewDumpOptions` was not.

## Tests to add

- A test that asserts, for every `OutputFormat` and every `CompressionType` constant, that a dump either writes a file or refuses with `ErrUnsupportedFormat`, and that the set that writes is exactly the set the godoc lists. `dump_options_test.go` is the file for it. Reading the godoc from the source at test time is what keeps the list honest: a constant added later fails the test until the list names it.
- The sibling case: `DumpOptions.FileExtension` should agree with the name of the file a dump writes under the same options, for each format and codec pair.

### LINKED ISSUE #1054: AddPath and AddFS say they load three formats; they load every format the package reads
Found at v0.56.0 (73d9e60) by loading a directory and an embedded filesystem holding one file of each format and comparing the tables against what the two methods promise.

## Phenomenon

`AddPath` says a directory "loads all CSV/TSV/LTSV files" and `AddFS` says it "Automatically finds all CSV, TSV, and LTSV files in the filesystem". Both load every format the package reads, which is CSV, TSV, LTSV, JSON, JSONL, Parquet, XLSX, ACH and Fedwire. The direction of the error is the one that costs a caller something: a reader is told that a directory holding JSON and Parquet beside CSV will contribute only the CSV, so they either write a separate step to handle the rest or filter the directory before loading it, and neither is needed. The package's own summary two lines up in `doc.go` names all seven text and container formats, so the two methods are the only place that says three.

## Reproduction

```go
dir := t.TempDir() // holding a.csv, b.tsv, c.ltsv, d.json, e.jsonl
db, err := filesql.Open(context.Background(), dir)
// then: SELECT name FROM sqlite_master WHERE type='table' ORDER BY name
```

```
AddPath godoc says a directory loads all CSV/TSV/LTSV files; the tables are [a b c d e]
```

```go
fsys := fstest.MapFS{
    "data/a.csv": ..., "data/b.tsv": ..., "data/c.ltsv": ..., "data/d.json": ..., "data/e.jsonl": ...,
}
db, err := filesql.NewBuilder().AddFS(fsys).Open(context.Background())
```

```
AddFS godoc says it finds all CSV, TSV, and LTSV files; the tables are [a b c d e]
```

Five files of five formats produce five tables through each door.

## Expected behavior

Both say what they load, which is every format this package reads, and neither enumerates a subset. `AddPath`'s "Directory:" example line and `AddFS`'s first sentence are the two places to change; `doc.go` already states the list and should stay the one place it is written out, so the two methods should point at the formats rather than repeat them.

## Cause (guess)

Both sentences date from when the package read three formats. Each format added since was given a `FileType` constant, a reader and a line in `doc.go`, and neither method's godoc was revisited, because nothing connects them to the enumeration.

## Tests to add

- A test that loads a directory holding one file of every format the package reads and asserts one table per file, and the same through `AddFS` with an `fstest.MapFS`. `builder_test.go` is the file for it. It fails when a format is added and the directory walk does not pick it up, which is the behavior the sentence is about.
- The sibling case that the same test should carry: a format the package does not read, in the same directory, contributes no table and does not fail the load.

### LINKED ISSUE #1055: EnableAutoSave's godoc sends a reader to Build, a method the package no longer has
Found at v0.56.0 (73d9e60) while reading `EnableAutoSave`'s godoc against the exported API.

## Phenomenon

`EnableAutoSave`'s godoc says that a source the package reads but cannot write back "is refused by Build rather than written as something else". There is no `Build`. `DBBuilder.Build` was deprecated in v0.52.0 and removed before v0.55.0, so a reader who wants to know when that refusal arrives is sent to a method they cannot call and cannot find in the documentation they are reading. The refusal is real and arrives from the terminal method: `Open`, `OpenReadOnly`, `LoadInto` or `LoadIntoTx`.

## Reproduction

```bash
go doc github.com/nao1215/filesql DBBuilder | grep Build
```

```
(no output: the type has no Build method)
```

```go
    AddPath("d.json").
    EnableAutoSave("").
    Open(context.Background())
```

```
filesql: unsupported file format: /tmp/.../d.json cannot be written back because this package reads that format but does not write it; save to a directory with a format it writes instead
```

The refusal comes from `Open`, which is where a reader has to look for it.

## Expected behavior

The sentence names the step that refuses, which is the terminal method the caller calls. Nothing else about the behavior changes.

## Cause (guess)

The sentence was written while `Build` was still the step that validated a builder. The removal updated the code and the CHANGELOG; this sentence, in another file, was not part of either.

## Tests to add

- A test that fails when the documentation of an exported symbol names an identifier this package no longer exports: collect the exported names from the package's own AST, collect the identifiers the doc comments name in backticks or as bare words matching an exported spelling, and report a name that is documented and gone. `doc_sync_test.go` is the file for it. A narrow version -- no doc comment mentions `Build` -- would pass forever without catching the next removal, so the general form is what is worth writing.
- A test that a source in a format the package cannot write back is refused by each of the four terminal methods, with the same sentinel, so the sentence's claim about when the refusal arrives is pinned rather than only its wording. `autosave_test.go` is the file for that one.

### LINKED ISSUE #1056: Open's godoc states a table-naming rule the package does not follow
Found at v0.56.0 (73d9e60) by loading files whose names hold characters an identifier cannot and comparing the table against what `Open`'s godoc says the name becomes.

## Phenomenon

`Open`'s godoc says a file becomes "a table named after it, with the extensions dropped and characters SQL cannot hold in a bare identifier replaced by underscores". Only three characters are replaced -- whitespace, a hyphen and a dot -- and every other character an identifier cannot hold is dropped, which is what `doc.go`'s "# Table Naming" section says. A caller who follows `Open`'s sentence writes the wrong table name: for `a!b.csv` they expect `a_b` and the table is `ab`, so the query fails with "no such table" and nothing points at the naming rule as the reason. The two statements are in the same package and the one a caller reads first is the wrong one, since `Open` is where they start.

## Reproduction

```go
for _, name := range []string{"user-data.csv", "my file.csv", "a!b.csv", "a@b.csv", "a(b).csv", "a+b.csv", "a%b.csv"} {
    // write the file, filesql.Open it, read the table name out of sqlite_master
}
```

```
user-data.csv    -> table "user_data"
my file.csv      -> table "my_file"
a!b.csv          -> table "ab"
a@b.csv          -> table "ab"
a(b).csv         -> table "ab"
a+b.csv          -> table "ab"
a%b.csv          -> table "ab"
```

The first two are the examples `Open`'s godoc gives, and they are the two the rule it states happens to cover. Every other character is dropped rather than replaced, and the last five names all collide on one table.

## Expected behavior

`Open`'s godoc states the rule the package follows, or points at the "# Table Naming" section rather than restating it. The rule itself is not in question: dropping is what makes `売上 2026.csv` become `売上_2026` rather than a name full of underscores, and it is documented in full one section away.

The collision the reproduction shows is a separate matter and behaves correctly: two such files in one directory are refused with `ErrDuplicateTable`.

## Cause (guess)

`Open`'s summary predates the table-naming section, which was written out in full later. Nothing connects the two, so the summary kept a rule that was a reasonable guess at the time.

## Tests to add

- A test that loads a file for each character class the rule distinguishes -- a hyphen, a dot, a space, an ideographic space, a punctuation character, a symbol, a combining mark, a leading digit, a name that keeps nothing -- and asserts the table name, in `table_test.go` beside the existing naming tests. The value is that the rule then has one executable statement of itself, and the two prose copies can be reduced to one.
- `doc_sync_test.go` already has `TestPackageDocMatchesTheColumnNameRule`, which holds `doc.go` to the column-naming code. The same shape of test for the table-naming rule would catch a second prose copy drifting from the first.

### LINKED ISSUE #1057: The Encoding type says the read side understands every encoding a save writes; three of six are refused
Found at v0.56.0 (73d9e60) by writing a table in each encoding a save accepts and loading the result back.

## Phenomenon

The `Encoding` type's godoc says "The read side already understands these encodings; without a matching write side a caller that decoded a legacy source before loading had no way to get one back". Three of the six are refused on load: Shift-JIS, EUC-JP and ISO-2022-JP. `WithEncoding`, one screen away, says the opposite and says it correctly -- "filesql reads UTF-8 only, so a file written in another encoding is for other tools rather than for loading back" -- so the same package tells a caller both things about the same six values. A caller who believes the type's sentence writes a Shift-JIS export expecting to load it back, and finds out at load time.

## Reproduction

```go
db, _ := filesql.Open(ctx, "d.csv") // holding 名前,値 / 東京,1
for each Encoding:
    filesql.DumpDatabase(ctx, db, dir, filesql.NewDumpOptions().WithEncoding(enc))
    filesql.Open(ctx, filepath.Join(dir, "d.csv"))
```

```
UTF8       written and loaded back: err=<nil> rows=[[東京 1]]
ShiftJIS   written and loaded back: err=.../d.csv: failed to read CSV record: filesql: invalid UTF-8: byte 0x96 at offset 0 is not part of a valid character rows=[]
EUCJP      written and loaded back: err=.../d.csv: failed to read CSV record: filesql: invalid UTF-8: byte 0xC1 at offset 2 is not part of a valid character rows=[]
ISO2022JP  written and loaded back: err=.../d.csv: filesql: parsing failed: failed to read CSV header: filesql: text encoding failed: input looks like ISO-2022-JP; filesql reads UTF-8, so transcode it first rows=[]
UTF16LE    written and loaded back: err=<nil> rows=[[東京 1]]
UTF16BE    written and loaded back: err=<nil> rows=[[東京 1]]
```

The two UTF-16 encodings load because they are written with a byte-order mark, which the read side decodes. The three legacy Japanese encodings carry no such mark and are refused, which is the documented and correct behavior of the read side.

## Expected behavior

The `Encoding` type says what `WithEncoding` says: this package reads UTF-8, and a file written in another encoding is for other tools. If the asymmetry is worth stating on the type -- UTF-16 loads back because of the mark, the other three do not -- it should be stated once, in one of the two places, with the other pointing at it.

## Cause (guess)

The type's godoc explains why the option exists, and its account of the motivating bug ("an in-place save changed the file's encoding on disk") is right; the clause about the read side was a compression of "the read side already understands byte-order marks and the codecs it can detect", which is true of UTF-16 and not of the three that need to be told.

## Tests to add

- A round-trip test over all six encodings that asserts which of them a load reads back and which it refuses, with the sentinel each refusal carries. `save_encoding_test.go` is the file for it. Pinning the refusals is what makes the documentation checkable: today the asymmetry is behavior nothing states.
- The sibling case worth carrying beside it: a value the target encoding cannot write fails the save with `ErrEncoding` and leaves the destination as it was, which `WithEncoding` promises and which the same test file can hold for each of the six.

### LINKED ISSUE #1058: internal/parser's documentation still offers the package to outside callers and points at streaming APIs it has none of
Found at v0.56.0 (73d9e60) while reading `internal/parser/doc.go` against the package it now describes.

## Phenomenon

`internal/parser/doc.go` still describes a published package. It says "This package can be used by filesql, prep, or any application that needs to parse tabular data files", and the package moved under `internal/` in v0.55.0, where no application outside this module can import it. Its "Memory Considerations" section then advises "For files larger than available memory, consider: Using streaming APIs for CSV/TSV", and there are no streaming APIs here to use: the chunked readers live in `internal/reader`, and this package's `Parse` reads a whole table into a `TableData` by design, which the same section says two paragraphs earlier. Both sentences are addressed to a caller who cannot exist, and the second sends the only reader who does reach this file -- a maintainer -- looking for an API in the wrong package.

## Reproduction

```bash
go doc github.com/nao1215/filesql/internal/parser | head -20
```

```
package parser // import "github.com/nao1215/filesql/internal/parser"
...
This package can be used by filesql, prep, or any application that needs to
parse tabular data files.
...
For files larger than available memory, consider:
  - Using streaming APIs for CSV/TSV
```

```bash
grep -rn "func .*Stream\|func .*Chunk" internal/parser/*.go | grep -v _test.go
```

```
(no output)
```

The advice names an API the package does not have. `go list` confirms the package is internal, so the "any application" sentence names a set with one member, this module.

## Expected behavior

The package documentation describes the package as it is: the reading side's whole-table parser, used by `filesql` and `prep` inside this module, with the streaming path pointed at `internal/reader` rather than at an API this package does not have.

## Cause (guess)

The move in v0.55.0 changed the import path and the package's visibility and left its doc comment as it was, since nothing about a doc comment fails to compile. The streaming advice is older still and predates `internal/reader`.

## Tests to add

There is no behavior here to pin, so the test is about the class rather than the case:

- Extend the check proposed for the stale snippets so that it also fails when a doc comment inside an `internal/` package tells an outside caller how to use it -- concretely, when it names an import path under `internal/` as something a caller may import. That is narrow enough to be worth having and broad enough to catch the next package that moves.
- Alternatively, and more simply: a test that every package's doc comment names only identifiers and packages that exist, which is the same check the snippet test needs and would catch "streaming APIs" only if the phrase named one. If that is too weak to be worth writing, say so in the fix and leave this case to review.

### LINKED ISSUE #1059: Three links in README's format table point at headings the file does not have
Found at v0.56.0 (73d9e60) by resolving every relative link in README.md against the headings the file has.

## Phenomenon

Three links in README's "Supported File Formats" table point at headings that are not in the file. The XLSX row sends a reader to "Excel sheet visibility" and the ACH and Fedwire rows send them to "ACH and Fedwire", twice; none of the three anchors exists, so on GitHub each one scrolls nowhere and on a rendered copy each is a dead link in the one table a reader consults to find out what the package does with their file. The material the three point at now lives in the godoc, which is where README's "Behavior and limits" section says the rules are.

## Reproduction

```bash
grep -n '^#' README.md
```

```
1:# filesql
18:## Try it in 30 seconds
52:## Why filesql?
61:## Features
72:## Installation
89:## Recipes
93:### Query files with SQLite
146:### Query with another SQL dialect
208:### Load into a database you already own
238:### Clean rows before loading with prep
298:## Supported File Formats
347:## Behavior and limits
351:## Examples
353:### API example index
357:#### filesql
387:#### prep
415:#### dialect
426:### Integration examples
441:## Related Projects
448:## Contributing
452:## Support
460:## License
464:## Contributors ✨
```

```bash
grep -n 'Excel sheet visibility\|ACH and Fedwire' README.md
```

```
308:| `.xlsx` | Excel | One table per sheet ... see [Excel sheet visibility](#excel-sheet-visibility) |
309:| `.ach` | ACH (NACHA) | One table per record kind; see [ACH and Fedwire](#ach-and-fedwire) |
310:| `.fed` | Fedwire | One message becomes one row 326 columns wide; see [ACH and Fedwire](#ach-and-fedwire) |
```

No heading in the file produces either anchor. The other 55 relative links in README.md resolve.

## Expected behavior

Each link points at something that exists. The sections it named moved to the package documentation, so the three should point there -- at the "# Excel Sheet Visibility" and "# ACH and Fedwire" sections of the godoc on pkg.go.dev -- which is also the direction README's own "Behavior and limits" section sends a reader for rules of this kind.

## Cause (guess)

The two sections were prose in README and were moved into `doc.go` when the rules were consolidated there; the move took the headings and left the links that named them.

## Tests to add

- A test that resolves every relative link in README.md: a link to a file has to name a file that exists, and a link to an anchor has to name a heading the file holds, with the anchor derived the way GitHub derives one (lowercase, spaces to hyphens, punctuation dropped). `doc_sync_test.go` is the file for it, beside the tests that already read README. This catches the class rather than the case: the next section that moves to godoc leaves the same debris.

### LINKED ISSUE #1060: CONTRIBUTING says the CI matrix pins two Go versions; it pins three
Found at v0.56.0 (73d9e60) by reading CONTRIBUTING.md against the workflow it describes.

## Phenomenon

CONTRIBUTING.md says the unit-test matrix pins two Go versions; it pins three. A contributor reading the paragraph believes a change has to build under 1.25.13 and 1.26.6, and a failure under 1.27.0 then arrives as a surprise from a version the document told them was not tested.

## Reproduction

```bash
grep -n "pins exactly those two" CONTRIBUTING.md
sed -n '29,35p' .github/workflows/unit_test.yml
```

```
30:...GO-2026-5972 (encoding/asn1) landed, and the CI matrix pins exactly those two.
        go:
          - "1.25.13"
          - "1.26.6"
          - "1.27.0"
```

The workflow's own comment above the list says 1.27.0 is there so that "nothing here would notice a new Go changing behavior this module depends on".

## Expected behavior

The sentence names what the matrix pins. The two versions it names are the two `govulncheck.yml` pins, which is where the sentence's reasoning about the advisories belongs; the unit-test matrix pins a third for a different reason and the paragraph should say so or stop naming a count.

## Cause (guess)

1.27.0 was added to the unit-test matrix when it was released, and the paragraph in CONTRIBUTING.md, which is about advisories rather than about the matrix, was not revisited.

## Tests to add

- Extend `release_metadata_test.go`, which already reads repository metadata, so that the Go versions CONTRIBUTING.md names are the versions the workflows pin. Reading them out of the YAML rather than restating them is what keeps the two from drifting again.

### LINKED ISSUE #1061: SECURITY.md says the Parquet allocation issue is open; it was closed a week ago
Found at v0.56.0 (73d9e60) by checking the state of every issue SECURITY.md links to.

## Phenomenon

SECURITY.md says a known limitation "is tracked as #526 and is open. It is not fixed". Issue #526 was closed as completed on 2026-08-27. A reader who follows the link to see what is being done finds a closed issue and has no way to tell whether the limitation was fixed and the document is stale, or the issue was closed for another reason and the limitation stands. It is the second: the closing comment records that the allocation is inside parquet-go and cannot be fixed from here, which is a decision rather than a fix.

## Reproduction

```bash
sed -n '72,78p' SECURITY.md
gh issue view 526 --repo nao1215/filesql --json number,state,stateReason
```

```
This is tracked as [#526](https://github.com/nao1215/filesql/issues/526) and is
open. It is not fixed, and forking the decoder...
{"number":526,"state":"CLOSED","stateReason":"COMPLETED"}
```

## Expected behavior

The paragraph says what is true: the limitation stands, the issue that tracked it is closed, and the reason it is closed is that the allocation is in a dependency this project does not fork. The advice to callers -- do not point a memory-constrained process at Parquet files you did not write -- is unchanged and is the part that matters.

## Cause (guess)

The issue was closed with a decision rather than a fix, and the security policy that cited it as open was not revisited, because nothing connects a document's prose to the state of the issues it links.

## Tests to add

- A test that reads the issue numbers SECURITY.md links to and fails when the document calls one open that GitHub reports closed would need network access from a unit test, which is not a trade worth making. Instead: a test asserting SECURITY.md does not describe the state of any issue -- it links to them and describes the limitation in its own words -- which is a rule that can be checked offline and removes the class rather than this instance. `doc_sync_test.go` is the file for it.

### LINKED ISSUE #1062: Six snippets in prep's documentation construct a processor with a package no caller can import
Found at v0.56.0 (73d9e60) by compiling the code blocks in prep's documentation against the API a caller outside this module has.

## Phenomenon

Six code snippets in prep's documentation construct a processor with `prep.NewProcessor(parser.CSV)`. That `parser` is `github.com/nao1215/filesql/internal/parser`, which no caller outside this module may import, and `NewProcessor` takes a `prep.FileType` rather than a `parser.FileType`. A reader copying any of them out of pkg.go.dev gets "use of internal package not allowed" from the compiler; `prep/doc.go:20` shows the right spelling, `prep.NewProcessor(prep.FileTypeCSV)`, so the package documents the call both ways. The prose at `prep/doc.go:81` names `parser.FileType` for the same reason.

## Reproduction

```bash
grep -rn 'NewProcessor(parser\.\|parser\.FileType' prep/
```

```
prep/doc.go:81:// neither has a parser.FileType to construct a Processor with
prep/doc.go:93://	processor := prep.NewProcessor(parser.CSV)
prep/processor.go:34://	processor := prep.NewProcessor(parser.CSV)
prep/processor.go:48://	processor := prep.NewProcessor(parser.CSV)
prep/processor.go:64://	processor := prep.NewProcessor(parser.CSV)
prep/processor.go:69://	processor := prep.NewProcessor(parser.CSV)
prep/processor.go:132://	processor := prep.NewProcessor(parser.CSV)
```

A program importing that path to write the documented call:

```
main.go:8:2: use of internal package github.com/nao1215/filesql/internal/parser not allowed
```

## Expected behavior

Every snippet says `prep.NewProcessor(prep.FileTypeCSV)`, which is what the constructor takes and what `prep/doc.go:20` already shows, and the prose names `prep.FileType`.

## Cause (guess)

v0.55.0 moved the parser under `internal/` and gave prep a `FileType` of its own with the same constants. The signature changed and the call sites in code were updated; the snippets, which nothing compiles, kept the old spelling.

## Tests to add

- The same check the root package needs for its own stale snippets: a test that extracts every code block from every doc comment in the module and fails when one names a package under `internal/`, or calls a function of this module with the wrong number of arguments. `doc_sync_test.go` is the file for it, and it should walk `prep` and `dialect` as well as the root package, since this is where the fault landed.

### LINKED ISSUE #1063: prep says a field and a header match when they differ only in case; the rule is snake_case conversion
Found at v0.56.0 (73d9e60) by matching a struct against headers that differ from the field name in the ways prep's documentation says are equivalent.

## Phenomenon

`prep/doc.go` says "A field and a header name the same column when they differ only in ASCII letter case, which is how filesql compares the column names it makes from those headers". That is not the rule. A field name is converted to snake_case and the result is compared, so `UserName` matches the header `user_name` -- which differs by an inserted underscore rather than by case -- and fails against `username`, `USERNAME` and even `UserName`, the header spelled exactly as the field is. A caller whose CSV header is `username` writes the field the documentation tells them to and gets `ErrUnknownColumn` naming a column their file does not have, with the message quoting a name (`user_name`) that appears in neither the struct nor the file.

## Reproduction

```go
type row struct{ UserName string }
for _, header := range []string{"user_name", "username", "USERNAME", "UserName"} {
    var out []row
    _, _, err := prep.NewProcessor(prep.FileTypeCSV).Process(strings.NewReader(header+"\nx\n"), &out)
    fmt.Printf("header %-10q: err=%v rows=%v\n", header, err, out)
}
```

```
header "user_name" : err=<nil> rows=[{x}]
header "username"  : err=struct field names a column the input does not have: UserName (column "user_name"); the input has [username] rows=[]
header "USERNAME"  : err=struct field names a column the input does not have: UserName (column "user_name"); the input has [USERNAME] rows=[]
header "UserName"  : err=struct field names a column the input does not have: UserName (column "user_name"); the input has [UserName] rows=[]
```

The last line is the one that shows the documented rule is not the implemented one: a header identical to the field name does not match it.

## Expected behavior

Two things have to be decided together, and this issue proposes the second:

1. The rule is snake_case conversion, so the documentation should say that: the field name is converted to snake_case and matched against the header, and a `name:"..."` tag overrides it. That is a documentation fix and changes no behavior.
2. Whether a header should also match when it differs from the converted name only in ASCII letter case -- `USER_NAME` against `UserName` -- is a real question, since a spreadsheet export commonly upper-cases its header row. Folding case on top of the conversion would accept it. Doing so cannot break a struct that works today, since a fold only widens what matches, unless two headers of a file fold together, which is already refused as a duplicate column.

## Cause (guess)

`prep/parser.go:94-97` compares `toSnakeCase(field.Name)` against the header. The doc sentence describes filesql's own column-name comparison, which does fold ASCII case, and was written as if prep used the same rule.

## Tests to add

- A table over field-name and header pairs, in `prep/parser_test.go`: `UserName`/`user_name` matches, `UserName`/`UserName` and `UserName`/`username` state whichever answer is chosen, and the `name:"..."` tag matches whatever it names. The point of the table is that the rule then has one executable statement of itself.
- The failing case is worth its own assertion: the error message quotes the converted name, so a test should hold it to naming both the field and the name it looked for, which is what makes the failure diagnosable.

### LINKED ISSUE #1064: prep reads a name struct tag it documents nowhere, and FileTypeUnsupported names a function prep has not got
Found at v0.56.0 (73d9e60) by reading prep's exported documentation against the tags its parser reads.

## Phenomenon

`prep` reads a third struct tag, `name`, and documents neither its existence nor what it does. `prep/doc.go` says "It uses struct tags for validation ("validate" tag) and preprocessing ("prep" tag)", and `go doc -all ./prep` never spells `name`. The tag is read at `prep/parser.go:94` and overrides the column a field matches, which makes it the only way to map a field whose Go name does not convert to the header the file carries -- the case a caller meets as soon as their header is `username` rather than `user_name`. So the package holds the answer to its most common matching failure and does not mention it.

Beside that, `prep.FileTypeUnsupported`'s godoc says "It is what DetectFileType answers for a name it does not recognize", and there is no `DetectFileType` in `prep` or in `filesql`: the function lives in `internal/parser`, where a caller cannot reach it. A reader is sent to a call they cannot make to find out what the constant means.

## Reproduction

```go
type row struct {
    Whatever string `name:"user_name"`
}
var out []row
_, _, err := prep.NewProcessor(prep.FileTypeCSV).Process(strings.NewReader("user_name\nq\n"), &out)
fmt.Printf("err=%v rows=%v\n", err, out)
```

```
err=<nil> rows=[{q}]
```

```bash
go doc -all ./prep | grep -c 'name:"'
go doc ./prep DetectFileType
go doc . DetectFileType
```

```
0
doc: no symbol DetectFileType in package ./prep
doc: no symbol DetectFileType in package .
```

## Expected behavior

`prep/doc.go` names all three tags where it names two, and says what `name` does: it says which column a field matches, in place of the name derived from the field. `FileTypeUnsupported` says what it is without naming a function the package does not have.

## Cause (guess)

The `name` tag was added with the column matching and never reached the documentation, which lists the two tags the package is about. The `DetectFileType` sentence predates the parser package moving under `internal/`, where the function it names went with it.

## Tests to add

- A test in `prep/parser_test.go` that a field tagged `name:"..."` matches that column and that an untagged field matches the converted name, so the tag has an executable statement of itself rather than only a mention.
- The doc-comment check the module needs anyway -- no doc comment names an identifier the module does not export -- would have caught the `DetectFileType` sentence. `doc_sync_test.go` is the file for it.

### LINKED ISSUE #1065: prep:"default" with no parameter is accepted under WithStrictTagParsing, and silences a missing column
Found at v0.56.0 (73d9e60) by writing every parameterised prep tag with no parameter under WithStrictTagParsing.

## Phenomenon

`prep/doc.go` says "A tag that needs a parameter and is given none is an invalid tag argument: WithStrictTagParsing reports it, and without that option it is ignored." Eleven of the twelve parameterised tags do that. `prep:"default"` with no parameter is accepted in silence under strict parsing, and what it then does is not nothing: a bare `default` satisfies the escape that `doc.go` spells `prep:"default=..."`, so a field naming a column the input does not have stops being `ErrUnknownColumn` and quietly loads as the zero value. A caller who typed `prep:"default"` where they meant `prep:"default="` gets no complaint under the option whose whole purpose is to complain, and a missing column stops being reported.

## Reproduction

```go
type row struct {
    Name string
    Nope string `prep:"default"`
}
var out []row
_, _, err := prep.NewProcessor(prep.FileTypeCSV, prep.WithStrictTagParsing()).
    Process(strings.NewReader("name\nx\n"), &out)
fmt.Printf("err=%v rows=%v\n", err, out)
```

```
err=<nil> rows=[{x }]
```

The same shape for the other tags, all under `WithStrictTagParsing`:

```
strict prep:"default"  -> err=<nil>
strict prep:"nullify"  -> err=field V: invalid tag format: nullify requires a value
strict prep:"prefix"   -> err=field V: invalid tag format: prefix requires a value
strict prep:"replace"  -> err=field V: invalid tag format: replace requires old:new format, got ""
strict prep:"truncate" -> err=field V: invalid tag format: truncate requires a positive integer, got ""
strict prep:"coerce"   -> err=field V: invalid tag format: coerce requires int, float, or bool, got ""
```

## Expected behavior

One of two, and the first is what the documentation already promises:

1. `prep:"default"` with no parameter is an invalid tag argument, reported under `WithStrictTagParsing` like the other eleven and ignored without it. A caller who means "the empty string when the column is missing" writes `prep:"default="`, which is unambiguous and is what the documentation shows.
2. Or a bare `default` means `default=`, in which case the documentation says so beside the sentence that says every parameterised tag needs its parameter, because it is the exception to it.

The exemption exists today only in an unexported comment at `prep/parser.go:204-206`, so neither reading is available to a caller.

## Cause (guess)

`default` was exempted from the parameter check deliberately -- the comment says so -- because an empty default is a meaningful value. The blanket sentence in `doc.go` was written afterwards and does not carry the exception.

## Tests to add

- A table in `prep/parser_test.go` over every parameterised tag written with no parameter, under strict parsing and without it, asserting which report and which are ignored. Whichever answer is chosen for `default`, the table is what keeps the twelfth from drifting from the eleven again.
- The consequence is the case worth pinning separately: a field naming a column the input does not have, tagged with a bare `default`, either reports `ErrUnknownColumn` or loads the zero value -- and the test says which, because that is the behavior a caller depends on.

### LINKED ISSUE #1066: prep refuses a duplicate header with an error matching no sentinel the packages document
Found at v0.56.0 (73d9e60) by giving prep a header that names one column twice and matching the error against the sentinels the packages document.

## Phenomenon

A header naming one column twice is refused by `prep` with a bare error that matches none of the package's sentinels and none of filesql's. `prep/errors.go` heads its sentinel block with "Sentinel errors this package returns", and the root `doc.go` says a header that names one column twice "is refused with ErrDuplicateColumn" -- prep reads its input through the same parser, so a caller who branches on that sentinel to tell a duplicate header from any other failure takes the wrong branch through prep. The message is right and only the error's identity is missing, which is the difference between a caller being able to handle the case and having to match on the text.

## Reproduction

```go
var out []struct{ A string }
_, _, err := prep.NewProcessor(prep.FileTypeCSV).Process(strings.NewReader("a,a\nx,y\n"), &out)
fmt.Printf("err=%v\n", err)
fmt.Printf("errors.Is(err, filesql.ErrDuplicateColumn) = %v\n", errors.Is(err, filesql.ErrDuplicateColumn))
```

```
err=duplicate column name: "a" (column 2)
errors.Is(err, filesql.ErrDuplicateColumn) = false
```

The same input through `filesql.Open` matches `ErrDuplicateColumn`, so one document produces two different kinds of failure depending on which door it goes through.

## Expected behavior

The refusal carries `filesql.ErrDuplicateColumn`, which prep already re-exports the way it re-exports the rest of the read side's sentinels, so `errors.Is` finds it through either door. The message stays as it is: it names the column and its position, which is what a caller shows a user.

## Cause (guess)

The refusal is a bare `fmt.Errorf` at `internal/parser/parser.go:286`. The loader wraps what the parser returns in its own error, which is where the sentinel is attached, and prep returns the parser's error directly.

## Tests to add

- A test in `prep/processor_test.go` that the refusal matches `ErrDuplicateColumn`, for CSV and for LTSV, since LTSV carries its labels per record and refuses the same thing at a different point.
- The broader check worth having beside it: every error `prep.Process` returns for a malformed input matches one of the sentinels the package documents. A test that walks a table of bad inputs and asserts each error matches at least one exported sentinel would have found this one, and finds the next.

### LINKED ISSUE #1067: Four constructs reach SQLite untranslated and fail as unknown names, which the dialect documentation says never happens
Found at v0.56.0 (73d9e60) by running the constructs dialect/doc.go says are never handed to SQLite.

## Phenomenon

`dialect/doc.go` says translation refuses by name what it cannot translate and "never" hands a name to SQLite that only the source dialect knows, because "no such function" tells a caller their own spelling does not exist, which is a worse answer than saying the construct has no SQLite form. Two families still reach SQLite that way. A PostgreSQL `regexp_match` -- which `doc.go` itself names as not implemented -- and a MySQL `MASTER_POS_WAIT` are passed through and fail as unknown functions. PostgreSQL's `current_user`, `session_user` and `current_catalog` written as bare keywords, which is their standard spelling, are passed through and fail as unknown *columns*; the same three written with parentheses are refused here, by name and with the reason. So whether a caller is told what is wrong depends on whether they wrote the parentheses.

## Reproduction

```go
dialect.RegisterFunctions()
for _, q := range []struct{ d dialect.Dialect; sql string }{
    {dialect.PostgreSQL, "SELECT regexp_match('abc','b')"},
    {dialect.MySQL, "SELECT MASTER_POS_WAIT('a',1)"},
    {dialect.PostgreSQL, "SELECT current_user"},
    {dialect.PostgreSQL, "SELECT current_user()"},
} {
    translated, err := dialect.Translate(q.d, q.sql)
    if err != nil { fmt.Println("REFUSED:", err); continue }
    var v any
    fmt.Println(translated, "=>", db.QueryRow(translated).Scan(&v))
}
```

```
SELECT regexp_match('abc', 'b') AS "regexp_match('abc','b')" => SQL logic error: no such function: regexp_match (1)
SELECT MASTER_POS_WAIT('a', 1) AS "MASTER_POS_WAIT('a',1)" => SQL logic error: no such function: MASTER_POS_WAIT (1)
SELECT current_user => SQL logic error: no such column: current_user (1)
REFUSED: dialect: syntax not supported on SQLite backend: CURRENT_USER is not supported; it answers a fact about the server or its catalog, which a database made from files does not have at line 1, column 8
```

The last two lines are the same construct in its two spellings, answered two different ways.

## Expected behavior

All four are refused here, by name, with the reason the parenthesized spelling already gives. `regexp_match` and `regexp_matches` are refused because their answer is an array, which is the reason `doc.go` gives for not implementing them; `MASTER_POS_WAIT` is refused as one of the replication functions, which is the family the refusal tables already carry; and the three bare keywords are refused with the message their parenthesized spelling gets.

The bare keywords are the same shape as the clock keywords fixed in v0.56.0: a bare word that is not in the table of words standing for a value stays a column reference, and SQLite answers about a column rather than about the construct.

## Cause (guess)

`regexp_match`, `regexp_matches` and `MASTER_POS_WAIT` are missing from the refusal tables, which were swept from the engines' own function lists -- the sweep test in `dialect/testdata/engine_functions_*.txt` covers the names those engines report, and `untranslated_postgresql.txt` records the ones with no answer, so these three are either absent from the lists or listed as known debt. The three bare keywords are missing from the table of bare words the translator rewrites, the same table the clock keywords were added to.

## Tests to add

- The existing sweep that holds every engine function name to "translated or refused by name" should cover these; if `regexp_match` is in `untranslated_postgresql.txt` today, moving it out of that file is part of the fix, since that file is the record of what has no answer and a refusal is an answer.
- A table in `dialect/dialect_test.go` of every construct that has both a bare-keyword and a parenthesized spelling -- `current_user`, `session_user`, `current_catalog`, `current_schema`, `user` -- asserting that the two spellings answer the same way. That is the shape of the fault, and it is what the clock-keyword fix established for its own family.

### LINKED ISSUE #1068: Three statements in the dialect documentation describe behavior the package no longer has
Found at v0.56.0 (73d9e60) by running the constructs dialect/doc.go describes as refused or as not implemented.

## Phenomenon

Three statements in `dialect/doc.go` describe behavior the package no longer has, and each sends a reader the wrong way about a construct that works.

`doc.go` lists among the refusals "AT TIME ZONE and a timestamp typed WITH TIME ZONE, since SQLite keeps no zone with a timestamp". `AT TIME ZONE` is refused; a cast to `timestamp with time zone` or `timestamptz` translates and answers. A reader is told not to write a cast that works.

`doc.go` says "the functions whose answer is one of those types -- PostgreSQL's REGEXP_MATCH, SCALE and TRIM_SCALE ... -- are not implemented, since there would be no value to return". `SCALE` and `TRIM_SCALE` are implemented, registered as runtime functions, and answer what PostgreSQL answers. Only `REGEXP_MATCH` is missing, and its answer really is an array.

`doc.go` says what is refused includes "a value that is not the same twice". `UUID()` is refused with exactly that reason, and `RAND()`, `random()` and `GENERATE_UUID()` are translated and answer a different value each time. The clock functions have a documented exception for this; the random ones have none.

## Reproduction

```
postgresql SELECT CAST('2024-01-01 12:00:00' AS timestamp with time zone) -> postgresql_cast(...) = 2024-01-01 12:00:00 (err <nil>)
postgresql SELECT '2024-01-01 12:00:00'::timestamptz                      -> postgresql_cast(...) = 2024-01-01 12:00:00 (err <nil>)
postgresql SELECT SCALE(1.50)                                             -> SELECT SCALE(1.50)      = 1   (err <nil>)
postgresql SELECT TRIM_SCALE(1.50)                                        -> SELECT TRIM_SCALE(1.50) = 1.5 (err <nil>)
mysql      SELECT RAND()                                                  -> SELECT RAND()           = 0.6970433651785793 (err <nil>)
mysql      SELECT RAND()                                                  -> SELECT RAND()           = 0.27778616652404653 (err <nil>)
postgresql SELECT random()                                                -> postgresql_random()     = 0.4465743842001658 (err <nil>)
googlesql  SELECT GENERATE_UUID()                                         -> GENERATE_UUID()         = a2f76b6d-... (err <nil>)
mysql      SELECT UUID()                                                  REFUSED: UUID is not supported; its value is not the same twice
```

Each line is `dialect.Translate` followed by running the result through `modernc.org/sqlite`. `SCALE(1.50)` answering 1 and `TRIM_SCALE(1.50)` answering 1.5 are what PostgreSQL answers, so the implementations are right and the sentence about them is wrong.

## Expected behavior

The documentation says what the package does:

- A cast to a timestamp with a zone is translated and answers the value without a zone, since SQLite keeps none; `AT TIME ZONE`, which asks for a conversion between zones, stays refused. Whether the cast should instead be refused is a separate question, and translating it is the more useful answer: the value survives and only the zone is dropped, which is what every other zone-carrying type here does.
- `SCALE` and `TRIM_SCALE` are implemented. Only `REGEXP_MATCH` is not, and its reason is its own.
- The refusal of a value that is not the same twice is `UUID`'s alone, and the random functions are the exception: a caller asking for a random number is asking for a different answer each time, which is what they get, whereas `UUID()` was refused for a reason recorded elsewhere. State the exception where the category is stated, the way the clock functions' exception is stated.

## Cause (guess)

All three sentences were written when the categories were being drawn and each was overtaken by a later change: the cast lowering learned the zone-carrying types, the two decimal functions were implemented in the arithmetic batch, and the random functions were added without revisiting the category that would have covered them.

## Tests to add

- The refusal tables already have a test that holds every engine function name to "translated or refused by name". Extend it to assert the opposite direction for the names the documentation calls not implemented: a name `doc.go` says is not implemented must be refused, so the sentence cannot outlive the implementation again. `dialect/dialect_test.go` is the file for it.
- A test that pins the three answers this issue is about -- the timestamptz cast, `SCALE`/`TRIM_SCALE`, and that `RAND()` twice gives two values -- so the documentation's new sentences have executable statements of themselves.

### LINKED ISSUE #1069: A MySQL LTRIM with two arguments answers instead of being refused
Found at v0.56.0 (73d9e60) by calling a MySQL function with an argument count MySQL does not accept.

## Phenomenon

`dialect/doc.go` says "A call whose argument count no form of the function accepts is refused here too, since it would otherwise fail under the helper's name." A MySQL `LTRIM('xxab','x')` is not refused: it is passed to SQLite, whose `LTRIM` does take a second argument, and answers `ab`. MySQL's `LTRIM` takes one argument and answers an error, so a query written against MySQL and run here gets an answer MySQL would not give -- silently, which is the outcome the package's compatibility promise is written to avoid. `LTRIM` is not an obscure case: `doc.go` lists it among the calls this package supervises for MySQL's own text and number semantics.

The sibling case behaves correctly: `LOWER('a','b')` is refused by name, saying `LOWER` takes 1 argument and the call has 2.

## Reproduction

```go
dialect.RegisterFunctions()
translated, err := dialect.Translate(dialect.MySQL, "SELECT LTRIM('xxab','x')")
var v any
fmt.Println(translated, err, db.QueryRow(translated).Scan(&v), v)
```

```
SELECT LTRIM('xxab', 'x') AS "LTRIM('xxab','x')" <nil> <nil> ab
```

MySQL 8.4 answers `ERROR 1582 (42000): Incorrect parameter count in the call to native function 'LTRIM'` for the same query. `RTRIM` behaves the same way here.

## Expected behavior

`LTRIM` and `RTRIM` under MySQL take one argument, and a call with two is refused by name with the message the arity check already produces for `LOWER`. Under PostgreSQL and GoogleSQL both take two, so the check is per dialect, which is what the existing arity table is for.

## Cause (guess)

The arity check runs over the names the package rewrites; `LTRIM` and `RTRIM` are left on SQLite's own function for MySQL, so nothing checks their argument count and SQLite's two-argument form answers. The names that are rewritten are checked because the rewriting has to know how many arguments it is rewriting.

## Tests to add

- A table in `dialect/dialect_test.go` over every function name this package leaves on SQLite's own implementation, per dialect, asserting that a call with an argument count the source dialect does not accept is refused. The list of names is derivable from the lowering tables, so the test can be written over that rather than by hand, and it then covers the next name added to it.
- The two specific cases as regressions: `LTRIM('xxab','x')` and `RTRIM('abxx','x')` are refused under MySQL and answer under PostgreSQL and GoogleSQL.

nao1215:
## Summary

Nineteen findings from an audit of the documentation against the code: every claim in `doc.go`, in the godoc of each exported symbol, in README, CONTRIBUTING and SECURITY, and in the CI workflows, run against what the package actually does. Four are behavior bugs, the rest are documentation that stopped being true, mostly at the v0.55.0 breaking change.

The behavior fixes:

- The nightly fuzz workflow named `./parser`, which moved under `internal/` in v0.55.0, so `FuzzParseFormats` and `FuzzParseBinary` have failed with "directory not found" on every scheduled run since and have fuzzed nothing. The other five jobs passed beside them and a scheduled failure notifies no pull request, which is why it stood.
- PostgreSQL's `current_user`, `session_user` and `system_user` written without parentheses reached SQLite as column references and failed as unknown columns, while the parenthesized spelling was refused by name with the reason. Whether one is refused is now the dialect's own answer, read from the table that answers for the call, so the two spellings cannot say different things.
- A MySQL `LTRIM('xxab','x')` answered `ab`, because SQLite's `LTRIM` takes the second argument and MySQL's does not. MySQL answers `Incorrect parameter count`; it is refused here now.
- `prep` refused a duplicate header with an error matching no sentinel, although the root package documents `ErrDuplicateColumn` for the input and both doors read the document through the same reader.

The documentation fixes are in the changelog; the shape of them is that a signature changed, or a package moved, or a format was added, and the prose beside it was not part of what the compiler checked.

Closes #1051, closes #1052, closes #1053, closes #1054, closes #1055, closes #1056, closes #1057, closes #1058, closes #1059, closes #1060, closes #1061, closes #1062, closes #1063, closes #1064, closes #1065, closes #1066, closes #1067, closes #1068, closes #1069.

## Tests

The tests are written against the class rather than the case wherever the class could be checked:

- `doc_sync_test.go`: every code block in every doc comment is checked against the signatures the module exports, and no doc comment may name an exported symbol the module no longer has -- which found three references to `Build` beyond the one this started from. Every link in README resolves, an anchor against the headings the file holds and a path against the repository.
- `release_metadata_test.go`: every path a workflow names exists, and the fuzz matrix names every fuzz target in the module, in the package that holds it.
- `dump_options_test.go`: the format and compression lists are held to what a dump accepts, so a format added later cannot be missing from them in silence.
- `table_test.go`, `builder_test.go`, `save_encoding_test.go`, `prep/parser_test.go`, `prep/processor_test.go`, `dialect/dialect_test.go`: the rules the documentation states, each with the case that used to disagree.

Each test was run against the unfixed file to confirm it fails there.

## Verification

`go test ./...`, `make lint` (0 issues), `GOOS=windows go build ./...`, and sqly builds and passes its tests against this branch with a local `replace`. No exported name changes; `filesql.ErrDuplicateColumn` is the same sentinel with the same message, moved to the internal package both doors share.

## Not done here

`REGEXP_MATCH` and `MASTER_POS_WAIT` still reach SQLite: the first is recorded debt in `dialect/testdata/untranslated_postgresql.txt`, and the second is absent from the MySQL name list the refusal sweep was built from. Both are noted in #1067 rather than fixed, and the `doc.go` sentence that said this never happens now says what the tables promise. A GoogleSQL `BYTES` literal holding a high byte is read as UTF-8 text rather than as the byte it names; it is not filed yet and is worth its own issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL dialect validation for unsupported keywords and function argument counts, including MySQL `LTRIM` and `RTRIM`.
  * Duplicate-column errors are now consistently identifiable across loading and preparation workflows.
  * Restored nightly fuzz-testing coverage.

* **Documentation**
  * Clarified supported formats, encoding and compression behavior, table-name generation, dialect support, autosave validation, and preparation tags.
  * Updated API examples with context arguments and corrected documentation links.
  * Expanded guidance on security, CI setup, and known Parquet limitations.


nao1215:
All four review notes are addressed.

The duplicate `## [Unreleased]` heading was already merged in b66724e, before this review landed; there is one heading with the Fixed and Documentation sections under it.

The snippet checker's method lookup is fixed in 469cd71 and was a real hole: `moduleDocs` keyed a method as `package.Receiver.Name` and the lookup asked for `package.Name`, so every method call in a doc comment was passed over silently. A method is now found by its own name, ambiguous names are skipped rather than guessed at, and a qualifier is read as another package only where the file imports one by that name -- without that last part `os.Open("users.csv.gz")` resolved to this package's `Open` and was reported for not passing a context. Proved by breaking `builder.SetDefaultChunkSize(5000)` to `SetDefaultChunkSize()` and watching the test fail on it.

`prep:"default="` has a case of its own beside the bare spelling, asserting the same empty-string result.

The two `parser.Parse` references in `prep/doc.go` are gone. They are the same fault this pull request is about -- documentation naming a package that moved under `internal/` -- and I had missed them because they are prose rather than snippets; the checker that finds the snippets does not read prose.

PR #1072, opened 2026-09-03
fix(dialect): read a bytes literal as the bytes it names

### LINKED ISSUE #1071: A GoogleSQL BYTES literal with a byte above 127 becomes its UTF-8 encoding
Found at v0.56.0 (73d9e60) by asking GoogleSQL for the hexadecimal of a byte literal holding a byte above 127.

## Phenomenon

A GoogleSQL `BYTES` literal holding a byte above 127 is read as UTF-8 text rather than as the byte it names. `b'\xff'` becomes the two-byte blob `x'c3bf'` -- the UTF-8 encoding of U+00FF -- so `TO_HEX(b'\xff')` answers `c3bf` where BigQuery answers `ff`, and `BYTE_LENGTH(b'\xff')` answers 2 where BigQuery answers 1. The dialect documentation says "A literal that names bytes is read the way the dialect that wrote it reads it", and that a GoogleSQL `BYTES` becomes a BLOB. A caller comparing a hash or a key against a byte literal gets no match and no error.

## Reproduction

```go
dialect.RegisterFunctions()
for _, q := range []string{`SELECT TO_HEX(b'\xff')`, `SELECT BYTE_LENGTH(b'\xff')`, `SELECT TO_HEX(b'\x00\x7f\x80\xff')`} {
    translated, err := dialect.Translate(dialect.GoogleSQL, q)
    var v any
    fmt.Println(q, "->", translated, "=", v, db.QueryRow(translated).Scan(&v), err)
}
```

```
SELECT TO_HEX(b'\xff')             -> SELECT TO_HEX(x'c3bf')       = c3bf
SELECT BYTE_LENGTH(b'\xff')        -> SELECT octet_length(x'c3bf') = 2
SELECT TO_HEX(b'\x00\x7f\x80\xff') -> ...                          = 007fc280c3bf
```

The last is four bytes written and six read back. BigQuery answers `ff`, `1` and `007f80ff`.

## Expected behavior

`b'\xff'` is the one byte 0xFF, so `TO_HEX` answers `ff` and `BYTE_LENGTH` answers 1. A `BYTES` literal's escapes name bytes rather than code points, which is the difference between it and a string literal, and it is why BigQuery has two kinds of literal.

MySQL's and PostgreSQL's byte literals were checked in the same run and are read correctly: `0x41`, `x'41'`, `X'41'` and both bytea input formats all answer for the byte they name.

## Cause (guess)

The `b'...'` literal's escapes are almost certainly decoded into a Go string, where `\xff` becomes the rune U+00FF, and the string is then encoded as UTF-8 on its way into the blob. The decoding has to produce a byte slice for this literal, with `\xNN` writing one byte, and the two kinds of literal have to keep separate escape rules: the same escape in a GoogleSQL string literal does name a code point.

## Tests to add

- A table in `dialect/dialect_test.go` over a `BYTES` literal for each escape form -- `\xNN`, `\NNN` octal, `\\`, a literal high byte written raw -- asserting the hex the blob holds, running the translation through SQLite so the assertion is about the value rather than the spelling.
- The sibling that makes the rule visible: the same escapes in a GoogleSQL string literal name code points, so `'\xff'` is the character U+00FF and two bytes of UTF-8, which is the distinction being fixed.

nao1215:
## Summary

A GoogleSQL `BYTES` literal named its bytes and this package read them as characters. `b'\xff'` became the code point U+00FF and was stored as the two UTF-8 bytes that encode it, so `TO_HEX(b'\xff')` answered `c3bf` where BigQuery answers `ff`, `BYTE_LENGTH(b'\xff')` answered 2 where BigQuery answers 1, and `b'\x00\x7f\x80\xff'` reached SQLite as six bytes rather than four. A comparison against a hash or a key written that way matched nothing, with no error to say why.

The escapes of a `BYTES` literal name bytes and the escapes of a string literal name code points, which is what the two kinds of literal are for; both went through one decoder that produced a Go string, and `[]byte` of that string is its UTF-8. An escape naming a number now writes that byte in a literal that names bytes and the code point of that number in one that names characters. A raw byte written into the literal was already carried through unchanged, and `rb'...'`, which keeps its backslashes, is unaffected.

Closes #1071.

## Tests

`TestGoogleSQLBytesLiteralNamesBytes` in `dialect/dialect_test.go` covers each escape form -- `\xNN`, octal `\NNN`, `\\`, `\n`, a raw character, and the empty literal -- asserting the hexadecimal the blob holds, and carries the sibling that makes the rule visible: the same escape in a string literal is the character U+00FF. It fails on the three high-byte cases before the fix.

## Verification

`go test ./...`, `make lint` (0 issues), and sqly passes its tests against this branch with a local `replace`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected GoogleSQL `BYTES` literal handling so numeric escapes decode to raw byte values.
  * Preserved character-code-point decoding for regular string literals.
  * Maintained existing behavior for raw byte literals and `rb'...'` syntax.

* **Documentation**
  * Added changelog documentation describing the updated `BYTES` escape behavior.


nao1215:
Fixed. `b'\400'` through `b'\777'` are refused now, with the reason at the escape: "a byte string escape names a byte, and this one names a number past 255". The check reads the literal's source text rather than the decoded content, since that is where the caller's spelling is, and it steps over an escaped backslash so `b'\\400'` -- a backslash and three digits -- still loads.

The regression test covers `\377` (the last byte, accepted), `\400`, `\777` and a literal holding two such escapes, the escaped-backslash case, and the same escape in a string literal, where it names the code point 256 and is not refused because a string literal names code points.

PR #1080, opened 2026-09-03
fix: give each rule one implementation

### LINKED ISSUE #1073: prep refuses four classes of malformed input with no sentinel, because the kind-to-sentinel mapping is written twice
Found at v0.56.0 (66c3974) by running the same document through `filesql.Open` and through `prep.Process` and comparing the sentinels each refusal carries.

## Phenomenon

`prep` refuses four classes of malformed input with an error matching no sentinel at all, where a load refuses the same document with one. A caller who branches on the failure -- to tell "your file is malformed" from "your struct does not fit it" from "the file is not there" -- can do so through one door and not the other, although both read the document with the same reader. It is not four faults: `internal/reader.Kind` exists, in its own words, so that "a caller can report it with the sentinel error its own package already has", and the mapping from Kind to sentinel is written twice -- once in the loader, where all five kinds are handled, and once in `internal/parser.parseError`, where two are handled and the rest fall through a `default` that returns the error unclassified.

The four that reach a caller unclassified through prep are a record whose field count differs from the header, a column name holding a NUL, a syntax error such as an unterminated quote, and any other parse failure.

## Reproduction

```go
for _, input := range []string{
    "a,b\nx\n",           // a record with too few fields
    "a,b\nx,y,z\n",       // a record with too many fields
    "a\x00b,b\nx,y\n",    // a column name holding a NUL
    "a,b\n\"x,y\n",       // a field with an unterminated quote
} {
    db, loadErr := filesql.NewBuilder().AddReader(strings.NewReader(input), "d", filesql.FileTypeCSV).Open(ctx)
    if db != nil { db.Close() }
    var rows []struct{ A, B string }
    _, _, prepErr := prep.NewProcessor(prep.FileTypeCSV).Process(strings.NewReader(input), &rows)
    fmt.Printf("load: %v\nprep: %v\n\n", sentinelsOf(loadErr), sentinelsOf(prepErr))
}
```

```
load: refused[ErrColumnMismatch ErrParsing]: d: filesql: column count mismatch: row 1 has 1 fields, want 2
prep: refused[]: invalid CSV syntax: record on line 2 has 1 fields, the header has 2

load: refused[ErrColumnMismatch ErrParsing]: d: filesql: column count mismatch: row 1 has 3 fields, want 2
prep: refused[]: invalid CSV syntax: record on line 2 has 3 fields, the header has 2

load: refused[ErrInvalidData ErrParsing]: d: filesql: invalid data format: a column name cannot hold a NUL: column 1, "a\x00b"
prep: refused[]: a column name cannot hold a NUL: column 1, "a\x00b"

load: refused[ErrParsing]: d: filesql: parsing failed: failed to read CSV record: invalid CSV syntax: line 2: quoted field is never closed
prep: refused[]: failed to read CSV record: invalid CSV syntax: line 2: quoted field is never closed
```

The classes the two doors agree on are the ones `parseError` has a case for: a duplicate column, which was given one last week, and an empty input, which prep answers with its own `ErrEmptyFile`. Invalid UTF-8 agrees because that sentinel is attached deeper, in `internal/textin`, rather than by this mapping.

## Expected behavior

The mapping from a read failure's kind to a sentinel happens once, and a kind added later has to be given an answer rather than falling through a `default`. Both doors then classify the same document the same way, and prep's own answers -- `ErrEmptyFile` for an empty document, `ErrUnknownColumn` for a struct that does not fit -- stay prep's, since those are about prep rather than about the read.

Concretely: the read-side sentinels live where both doors can reach them, as `ErrDuplicateColumn` now does and as `ErrInvalidUTF8` and `ErrEncoding` already do in `internal/textin`; one function turns a `*reader.Error` into the error carrying its sentinel; and the loader and the parser both call it.

## Cause (guess)

The two switches were written at different times for different callers, and the `default` in the second is what let it stay incomplete: a switch over an enum whose purpose is to be mapped exhaustively should have no default that passes the value through. The loader's switch has the same shape but happens to be complete, which is why nothing showed the gap.

## Tests to add

- A table that runs one document through both doors and asserts they carry the same sentinel, over the inputs that reach each `Kind`: a duplicate header, a record whose field count differs, a column name holding a NUL, an unterminated quote, bytes that are not UTF-8, and an empty document, where prep's own answer is the documented difference. `prep/processor_test.go` is the file for it. The value is that the next kind added has a place that fails until both doors answer for it.
- A test over the `Kind` values themselves: every one has a sentinel, which is what makes the mapping exhaustive rather than the `default` making it look complete. It belongs beside the mapping.

### LINKED ISSUE #1074: An fs.FS holding a file named .ach fails the whole load, because two lists answer which files are supported
Found at v0.56.0 (66c3974) by handing the same tree to `AddFS` and to a directory on disk and comparing what each loaded.

## Phenomenon

An `fs.FS` holding a file named `.ach` or `.fed` -- the extension with no name before it -- fails the whole load with `filesql: unsupported file format: file type must be specified for reader input`, an error about a reader input the caller never created and a file type they were never asked for. Every other file in the filesystem is lost with it. The same tree in a directory on disk loads normally: the file is passed over and the tables beside it are read.

Which files the package loads is answered twice for an `fs.FS`: by a list of glob patterns and by the predicate the recursive walk uses. `.ach` matches the pattern `*.ach`, and the predicate refuses it, since `isACHFile` returns false for a path that is only the extension. The glob-matched file becomes a reader input with no file type, and the load fails on it. The two answers also sit one after the other in the same function, where the walk adds anything the predicate accepts and the glob adds anything a pattern matches, so the glob's only effect is to admit what the predicate would not.

## Reproduction

```go
fsys := fstest.MapFS{
    ".ach":   &fstest.MapFile{Data: []byte("a,b\n1,x\n")},
    "ok.csv": &fstest.MapFile{Data: []byte("a,b\n2,y\n")},
}
db, err := filesql.NewBuilder().AddFS(fsys).Open(ctx)
fmt.Println(err)
```

```
filesql: unsupported file format: file type must be specified for reader input
```

The same two files written into a directory:

```
a directory holding ".ach": err=<nil>, tables=[ok]
```

`.fed` behaves the same way. `.csv`, `.tsv` and `.json` load through both doors alike, since the predicate accepts those.

## Expected behavior

Both doors load the same tree. A file named `.ach` is not an ACH file by this package's own rule -- `isACHFile` says so -- so it is passed over, and the tables beside it load.

The fix worth making is not a special case for the two names: it is that "which files does this package load" has one answer. The walk over the filesystem already reaches every file the glob patterns could match, including the ones in subdirectories the patterns do not reach, and it asks the predicate; the glob pass above it adds nothing the walk would not add and admits what the walk refuses. Removing it leaves one answer, and `supportedFileExtPatterns` with it, which exists only for that pass.

## Cause (guess)

`file_processor.go` `processFSToReaders` globs `supportedFileExtPatterns()` and then walks the filesystem with `isSupportedFile`, deduplicating. The glob pass is older; the walk was added for subdirectories and covers the glob's ground. `supportedFileExtPatterns` builds its patterns from the extension constants directly, where `isSupportedFile` goes through `detectFileType`, which is where the "a path that is only the extension is not an ACH file" rule lives.

## Tests to add

- A test that hands the same tree to `AddFS` and to a directory and asserts the same tables and the same error, over a table of trees: one holding `.ach`, one holding `.fed`, one holding a file with no extension, one with a file in a deep subdirectory, one with two files sharing a base name, one with a compressed member beside its plain form. `file_processor_test.go` is the file for it. The value is the invariant rather than these names: the two doors are two implementations of one rule and this is what holds them together.
- If the glob pass is removed, a test that an `fs.FS` whose files are all in subdirectories still loads, since that is what the glob patterns did not reach and the walk does.

### LINKED ISSUE #1075: An in-place save with no file-backed source fails at Close rather than when the builder opens
Found at v0.56.0 (66c3974) by asking for an in-place save on a database whose sources have no files.

## Phenomenon

`EnableAutoSave("")` writes each table back to the file it was loaded from, and a database loaded only from readers or from an `fs.FS` has no such file for any table. That is knowable when the builder opens, and it is reported at `Close` instead: the session runs, the caller's edits are made, and `Close` answers `auto-save failed: no original paths available for overwrite`, by which point there is nothing to be done about it. The error carries no sentinel, so a caller cannot tell this failure from any other close failure without matching on its text.

The comparison that makes it a defect rather than a choice is one door along: a source whose format cannot be written back -- a JSON file, say -- is refused when the builder opens, with `ErrUnsupportedFormat` and a message saying to save to a directory instead. Both are the same mistake, a caller asking for an in-place save that cannot happen, and one is caught before the session and the other after it.

## Reproduction

```go
    AddReader(strings.NewReader("a,b\n1,x\n"), "d", filesql.FileTypeCSV).
    EnableAutoSave("").
    Open(ctx)
fmt.Println("open:", err)
db.Exec(`UPDATE d SET b = 'edited'`)
fmt.Println("close:", db.Close())
```

```
open: <nil>
close: auto-save failed: no original paths available for overwrite
```

Against the format case, which is the same shape of mistake:

```go
_, err := filesql.NewBuilder().AddPath("d.json").EnableAutoSave("").Open(ctx)
```

```
filesql: unsupported file format: .../d.json cannot be written back because this package reads that format but does not write it; save to a directory with a format it writes instead
errors.Is(err, filesql.ErrUnsupportedFormat) = true
```

A database holding one file-backed table and one reader-backed table behaves correctly and is not part of this: the file-backed table is written, the other is left unsaved, and `Close` answers nil, which is what `EnableAutoSave`'s godoc describes.

## Expected behavior

A builder configured for an in-place save with no source that has a file is refused when it opens, carrying a sentinel, the way the unwritable-format case is. The message should say the same thing that one says: pass an output directory.

`EnableAutoSave`'s godoc describes the per-table rule -- "a table created during the session has no file to be written back to and is left unsaved" -- and says nothing about a database where that is true of every table, which is the case this is about.

## Cause (guess)

The check that refuses an unwritable format runs over the paths the builder collected, in the builder's validation; a database with no paths has nothing for it to walk, so it passes. The "no original paths" error is raised by the save itself, at close, which is the only place that knows the set is empty.

## Tests to add

- A test that a builder with `EnableAutoSave("")` and only reader or `fs.FS` sources is refused when it opens, with the sentinel, for each of the four terminal methods, since all four validate the builder. `autosave_test.go` is the file for it.
- The sibling that keeps the fix from going too far: a database holding one file-backed source beside a reader-backed one opens, and its close writes the file-backed table and leaves the other, which is the documented behavior and must not become a refusal.

### LINKED ISSUE #1076: A call renamed onto a SQLite builtin fails under the builtin's name, because the arity table holds three of nineteen
Found at v0.56.0 (66c3974) by translating a call to every function name the engines report and asking SQLite to prepare each translation that succeeded.

## Phenomenon

`dialect/doc.go` says: "A call whose argument count no form of the function accepts is refused here too, since it would otherwise fail under the helper's name." That is what happens for three names today, and the table that exists to prevent it covers three of the nineteen names this package renames a call onto.

A caller writing `SELECT CHAR_LENGTH()` -- a mistake, but their own -- is told `wrong number of arguments to function length()`, naming a function they did not write and cannot find in their query. `CHARACTER_LENGTH` answers the same under MySQL and PostgreSQL, and PostgreSQL's `STRING_AGG()` answers `wrong number of arguments to function group_concat()`.

The check is `checkHelperArity`, which asks `helperTakesArgumentCount` for the arity of the name the call was renamed to. That reads two tables: `helperArity`, which the runtime's own registration fills, and `builtinArity`, a hand-written list of the SQLite builtins the lowerings rename onto. `builtinArity` holds `trim`, `octet_length` and `json_array_length`. The lowerings rename onto sixteen more, and a name in neither table is passed through unchecked, which is the `return true` at the end of the function.

## Reproduction

```go
dialect.RegisterFunctions()
db.Exec(`CREATE TABLE t (a TEXT)`)
for _, q := range []struct{ d dialect.Dialect; sql string }{
    {dialect.MySQL, "SELECT CHAR_LENGTH() FROM t"},
    {dialect.PostgreSQL, "SELECT CHARACTER_LENGTH() FROM t"},
    {dialect.PostgreSQL, "SELECT STRING_AGG() FROM t"},
} {
    translated, err := dialect.Translate(q.d, q.sql)
    if err != nil { fmt.Println("refused:", err); continue }
    _, prepErr := db.Prepare(translated)
    fmt.Printf("%s -> %s: %v\n", q.sql, translated, prepErr)
}
```

```
SELECT CHAR_LENGTH() FROM t       -> SELECT length() AS "CHAR_LENGTH()" FROM t:        wrong number of arguments to function length() (1)
SELECT CHARACTER_LENGTH() FROM t  -> SELECT length() AS "CHARACTER_LENGTH()" FROM t:   wrong number of arguments to function length() (1)
SELECT STRING_AGG() FROM t        -> SELECT group_concat() AS "STRING_AGG()" FROM t:   wrong number of arguments to function group_concat() (1)
```

The names the lowerings rename onto, from `rename(call, "...")` in `dialect/internal/lower`, that are SQLite's own functions rather than registered helpers, with the argument counts SQLite accepts, measured rather than guessed:

```
COUNT [0 1]        INSTR [2]           group_concat [1 2]   json [1]
json_patch [2]     json_quote [1]      length [1]           ln [1]
log [1 2]          ltrim [1 2]         rtrim [1 2]
json_array, json_extract, json_insert, json_object, json_set are variadic
trim [1 2], octet_length [1], json_array_length [1 2] are the three the table holds
```

## Expected behavior

A call renamed onto a SQLite builtin is checked against that builtin's arity, so the refusal names what the caller wrote, which is what `doc.go` promises and what `LOWER('a','b')` already gets.

The fix worth making is not eleven more table entries: it is that the table cannot be silently incomplete again. The rename targets are enumerable from the lowering source -- every one is written `rename(call, "name")` -- so a test can require each to be in one of the two tables, with the variadic ones listed as taking any count rather than left out. Then a lowering that renames onto a new builtin fails until its arity is written down.

## Cause (guess)

`builtinArity` was added with the three names a particular fix needed, and nothing connects it to the set of names the lowerings rename onto. `helperTakesArgumentCount` ends with `return true`, which is right for a name the caller wrote and wrong for one this package substituted: those are two cases and the function cannot tell them apart, since it is given only the lowered name. Passing what the caller wrote alongside would let it.

## Tests to add

- A test that reads `dialect/internal/lower/*.go` for every `rename(call, "...")` target and asserts each is in `helperArity` or `builtinArity`. That is the check that makes the table complete rather than long. `dialect/internal/lower/helper_names_test.go` is the file for it.
- A test that translates a call to every name in `dialect/testdata/engine_functions_*.txt`, at several argument counts, and asserts that a translation which succeeded prepares in SQLite or fails naming a function the caller wrote. It is the oracle that found these three, it runs over a corpus the repository already has, and it catches the class rather than the case. `dialect/dialect_test.go` is the file for it.
- The three cases as regressions: `CHAR_LENGTH()`, `CHARACTER_LENGTH()` and `STRING_AGG()` are refused here, naming themselves.

### LINKED ISSUE #1077: A transaction opened by a statement that is not first is invisible to auto-save, and Close reports success over discarded edits
Found at v0.56.0 (66c3974) by opening a transaction with a statement whose first word is not the one that opens it.

## Phenomenon

A caller's edits are discarded and `Close` reports success. `db.Exec("SELECT 1; BEGIN")` opens a transaction in SQLite, which runs every statement of the string it is given; the guard that watches for a transaction reads only the leading word of the string, sees `SELECT`, and records nothing. The session then behaves as if a transaction were open -- an `UPDATE` inside it is visible to the session -- and at `Close` the auto-save runs, because the guard believes there is no transaction, while the uncommitted work is discarded with the connection. The file keeps the rows it had, `Close` answers nil, and nothing says the caller's work is gone.

The same session with `BEGIN` alone behaves correctly: `Close` answers "auto-save was skipped because the database was closed while a transaction was still open; commit or roll back before closing", and the caller can act on it. So a transaction the guard sees is reported and one it does not is silently lost, and what decides which is whether the keyword happens to be first in the string.

`doc.go` states the promise this breaks: "Auto-save counts such a transaction as open and refuses to save while it is, the same as one begun with BeginTx, and saves at a COMMIT run as a statement just as it does at tx.Commit."

## Reproduction

```go
path := "d.csv" // holding: a,b\n1,original\n
db, _ := filesql.NewBuilder().AddPath(path).EnableAutoSave("").Open(ctx)
db.Exec("SELECT 1; BEGIN")            // or "SELECT 1; SAVEPOINT sp"
db.Exec("UPDATE d SET b = 'edited'")
var seen string
db.QueryRow("SELECT b FROM d").Scan(&seen)
fmt.Println("the session sees:", seen)
fmt.Println("close:", db.Close())
after, _ := os.ReadFile(path)
fmt.Printf("the file holds: %q\n", after)
```

```
the session sees: edited
close: <nil>
the file holds: "a,b\n1,original\n"
```

The same with `BEGIN` alone, which is the behavior this should have:

```
the session sees: edited
close: filesql: database operation failed: auto-save was skipped because the database was closed while a transaction was still open; commit or roll back before closing
the file holds: "a,b\n1,original\n"
```

`"SELECT 1;\nBEGIN"` and `"SELECT 1; SAVEPOINT sp"` behave the same way as the first. A `COMMIT` that is not the leading keyword is invisible in the same way, so a transaction the guard did record stays recorded after it has ended.

## Expected behavior

Either reading: a transaction opened by any statement of an `Exec` is counted, so the save is skipped and `Close` says so; or such a string is refused. What must not happen is a save that runs while the database has an open transaction and a `Close` that reports success over discarded work.

## Cause (guess)

`readTxStatement` in `transaction.go` reads the leading word of the query and answers for that one statement. `database/sql`'s `Exec` hands the whole string to the driver, and SQLite runs every statement in it, so the guard is reading one statement where the engine runs many. It is the same shape as the comment case that was fixed before -- the guard reading a statement more narrowly than SQLite does -- one level up: there the question was where the statement starts, here it is how many there are.

The fix is to read every statement of the string rather than the first: split on the semicolons that are not inside a string, an identifier quote or a comment, and apply the reading this function already does to each, taking the last effect as the state the string leaves behind. The machinery for skipping comments and quoted identifiers is next to it, in `skipToStatement` and `leadingIdentifier`.

## Tests to add

- A table in `transaction_test.go` over strings holding more than one statement: `SELECT 1; BEGIN`, `BEGIN; COMMIT`, `SELECT 1; SAVEPOINT sp`, `SAVEPOINT sp; RELEASE sp`, a semicolon inside a string literal (`SELECT ';BEGIN'`), one inside a quoted identifier, one inside a line comment and one inside a block comment. Each asserts the transaction state the guard records against what SQLite has, which is what a save reads.
- The end-to-end case in `autosave_test.go`: a session that opens a transaction with a statement that is not first, edits a row, and closes, asserts either that the file has the edit or that `Close` reports the save was skipped -- never that `Close` succeeds while the edit is gone.

### LINKED ISSUE #1078: A table name too long for a file name is refused by the operating system rather than by this package
Found at v0.56.0 (66c3974) by exporting a database holding a table whose name is longer than a file name may be.

## Phenomenon

`DumpDatabase` refuses a table name that cannot be a file name in this package's own words, listing which names those are, and says the set is the same everywhere: "The same set is refused on every platform, so a database dumped on Linux and on Windows agrees about which tables it can write." A name too long for the filesystem is not in that set. It reaches the operating system, which refuses it with `file name too long`, wrapped in `failed to inspect <path>: stat <path>:` -- a message about a stat rather than about the table -- and the limit that decides it is the platform's rather than this package's, so a database that exports on one system fails on another, which is what the sentence says does not happen.

A 300-character table name is refused on Linux, whose limit is 255 bytes for one component. Windows refuses at a different point and for a different reason, since its limit is on the path rather than the component.

## Reproduction

```go
db, _ := filesql.Open(ctx, "d.csv")
name := strings.Repeat("n", 300)
db.Exec(`CREATE TABLE "` + name + `" (x TEXT)`)
db.Exec(`INSERT INTO "` + name + `" VALUES ('v')`)
err := filesql.DumpDatabase(ctx, db, outDir)
fmt.Println(err)
```

```
filesql: I/O operation failed: failed to export table nnnn...: filesql: I/O operation failed: failed to inspect /tmp/.../out/nnnn....csv: stat /tmp/.../out/nnnn....csv: file name too long
```

Every other unusable name is refused before anything is written, in the package's words and with the advice to rename:

```
filesql: invalid data format: table "../escaped" cannot be dumped because its name is not usable as a file name inside /tmp/.../out
filesql: invalid data format: table "with space" cannot be dumped because a load of "with space.csv" would name it "with_space"; rename the table first
```

## Expected behavior

A name too long to be a file name is refused the way the rest of the set is: before the export begins, in this package's words, with the advice to rename the table, and at the same length on every platform so that the documented agreement holds. 255 bytes is the component limit on Linux and macOS and is the smaller of the two, so refusing above it is the choice that keeps a database exportable everywhere; the file extension and the compression suffix are part of what has to fit.

## Cause (guess)

The name check in `save.go` enumerates the characters and the reserved device names, which are the two things a name can be wrong in a way that is knowable from the name alone -- and length is a third, which was not enumerated. The failure then arrives from `os.Stat` in the code that decides whether the destination exists.

## Tests to add

- A test in `dump_test.go` that a table name longer than the limit is refused with `ErrInvalidData` and the advice to rename, and that one at the limit exports, with the extension and any compression suffix counted. The boundary is what makes the test worth having: an off-by-one there is a name that exports on the developer's machine and not on a user's.
- The same table should carry the other refusals, since they are one set: a path separator, a device name, a name a load would spell differently. They are tested today in separate places, and one table makes the set visible.

### LINKED ISSUE #1079: A table name of whitespace is accepted by AddReader, refused by a save, and filled in by ExcelSheetTableNames
Found at v0.56.0 (66c3974) by giving a table name of whitespace to each door that takes one.

## Phenomenon

Which table names a caller may give is answered in three places, and they disagree.

`DBBuilder.AddReader` refuses an empty table name with `ErrInvalidData: table name must be specified for reader input`, and accepts one made of spaces or a tab. The table is then named `"   "`, which no save can write: an export refuses it, because a load of `"   .csv"` would name the table `___`, so the caller has built a database they cannot dump from a call that looked like every other. An input *path* of blanks is refused with `ErrEmptyPath` -- the same string, at the door next to it, gets the opposite answer -- and so is an auto-save output directory of blanks.

`ExcelSheetTableNames` takes the same name, for the same reason: its godoc says "a workbook has no file name when it arrives through AddReader: pass the table name given there instead". Given the empty string it answers `[sheet_Sheet1 sheet_Sheet2]` rather than refusing, so a caller who uses it to check a workbook before loading -- which is what it is exported for, "a caller that wants to refuse such a workbook before loading anything" -- is told the names are fine and then has the load refuse the empty name.

## Reproduction

```go
for _, name := range []string{"", "   ", "\t"} {
    db, err := filesql.NewBuilder().AddReader(strings.NewReader("a\n1\n"), name, filesql.FileTypeCSV).Open(ctx)
    fmt.Printf("AddReader(%q): %v tables=%v\n", name, err, tablesOf(db))
    names, err := filesql.ExcelSheetTableNames(name, []string{"Sheet1", "Sheet2"})
    fmt.Printf("ExcelSheetTableNames(%q): %v %v\n", name, names, err)
}
```

```
AddReader(""):            filesql: invalid data format: table name must be specified for reader input   tables=[]
ExcelSheetTableNames(""): [sheet_Sheet1 sheet_Sheet2] <nil>
AddReader("   "):            <nil> tables=[   ]
ExcelSheetTableNames("   "): [____Sheet1 ____Sheet2] <nil>
AddReader("\t"):            <nil> tables=[	]
ExcelSheetTableNames("\t"): [__Sheet1 __Sheet2] <nil>
```

The table a blank name makes cannot be exported:

```
filesql: invalid data format: table "   " cannot be dumped because a load of "   .csv" would name it "___"; rename the table first
```

## Expected behavior

A table name that is empty or blank is refused at the door it is given to, the way a path of blanks already is, so the failure arrives before the session rather than at the save. `ExcelSheetTableNames` refuses the same names, since its whole purpose is to answer what a load would do before the load runs.

The rule is one rule -- which names this package can carry from a caller through to a file -- and it should have one implementation that the three doors call, rather than a validator on one door, a sanitizer on the second and a dump-time check on the third.

## Cause (guess)

`AddReader`'s validation checks for the empty string, which is what its message says, and not for a name that trims to nothing; the path validator checks the second because a path of blanks was a bug once and was fixed there alone. `ExcelSheetTableNames` sanitizes rather than validates: an empty base sanitizes to nothing and the "a name that keeps no characters becomes sheet" rule fills it in, which is right for a name derived from a file and wrong for one a caller typed.

## Tests to add

- A table in `builder_test.go` over the names a caller may give a reader -- empty, spaces, a tab, an ideographic space, a name that sanitizes to nothing, a valid name -- asserting which are refused and with which sentinel, run through `AddReader` and through `ExcelSheetTableNames` so the two doors are held to one answer.
- The end-to-end case in `dump_test.go`: a name accepted by `AddReader` can be dumped. That is the invariant the disagreement broke, and it is worth stating directly, since it is what a caller is really relying on.

nao1215:
## Summary

Seven fixes, most of them one defect showing through several doors. The hunt was for rules this package implements more than once, since a rule with two implementations is a bug family rather than a bug.

The one that is not a family, and the most serious: a transaction opened by a statement that is not the first in a query was invisible to the guard, so `db.Exec("SELECT 1; BEGIN")` followed by an edit and a `Close` discarded the edit and answered nil. The driver runs every statement of the string it is given; the guard read the leading word. The same session opened with a bare `BEGIN` is told the save was skipped, so what decided whether a caller heard about their lost work was whether the keyword happened to be first.

The families:

- The mapping from a read failure's kind to a sentinel was written twice, and the second copy handled two kinds of five. Four classes of malformed input reached a caller of `prep` matching no sentinel at all, although `filesql` answered one for the same document read through the same reader. A read failure now carries its own sentinel and there is nothing left to map.
- Which files a filesystem contributes was answered by a list of glob patterns and by the walk's predicate, and the glob admitted what the predicate refuses: a file named `.ach` failed the whole load. The rule they disagreed about was itself reading the whole path rather than the file's own name, so the same file was an ACH file through one door and not the other.
- How many arguments a call may have was answered by a helper table, a builtin table holding three of twenty rename targets, and SQLite. `CHAR_LENGTH()` failed as `wrong number of arguments to function length()`, naming a function nowhere in the caller's query.
- Which table names a caller may give was answered by `AddReader`'s validator, by `ExcelSheetTableNames`'s sanitizer and by the dump's check, and they disagreed about a name of whitespace: accepted, filled in, and refused.

And two refusals that arrived too late or in the wrong voice: an in-place save with no file-backed source was reported by `Close` rather than at open, and a table name too long for a file name was refused by the operating system rather than here.

Closes #1073, closes #1074, closes #1075, closes #1076, closes #1077, closes #1078, closes #1079.

## Tests

Each new test is written against the rule rather than the case, so the next instance fails rather than the fixed one:

- `dialect/internal/lower/helper_names_test.go`: every `rename(call, "...")` target in the lowering source has an arity. It found a twentieth target the fix had missed.
- `dialect/dialect_test.go`: every translation of every name the engines report either prepares in SQLite or fails naming what the caller wrote. It is the oracle that found the arity gap.
- `file_processor_test.go`: one tree loads the same through a directory and through an `fs.FS`.
- `prep/processor_test.go`: one document refused at either door carries the same sentinel, over the eight inputs that reach each kind.
- `transaction_test.go`: the statements of a query are read whole, including the semicolons that end nothing because they are inside a string, an identifier or a comment.
- `builder_test.go`, `autosave_test.go`, `dump_test.go`: the end-to-end cases, including the invariant a name check is really about -- a name `AddReader` accepts is one a dump can write.

## Verification

`go test ./...`, `make lint` (0 issues), and sqly builds and passes its tests against this branch with a local `replace`. No exported name changes: `ErrEmptyData`, `ErrInvalidData`, `ErrUnsupportedFormat`, `ErrParsing`, `ErrDuplicateColumn` and `ErrColumnMismatch` are the same values with the same messages, moved to the internal package both doors share.

## Behavior a caller can see

Beyond the fixes: `BEGIN; UPDATE ...; COMMIT` in one `Exec` now saves at the commit, which it never did; a directory or a filesystem holding a file named `.ach` or `.fed` loads the tables beside it rather than failing; and a `MySQL`/`PostgreSQL` `CHAR_LENGTH` or `STRING_AGG` written with the wrong argument count is refused here, naming itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected transaction tracking for multi-statement SQL containing `BEGIN`, `SAVEPOINT`, and `COMMIT`.
  * Made file discovery consistent across directory and filesystem sources.
  * Improved handling of ACH and Fedwire filenames.
  * Added validation for unsupported in-place saves, blank table names, empty workbook paths, and filenames exceeding system limits.
  * Standardized parsing and validation errors so failure types can be identified reliably.
  * Improved SQL function argument-count validation across supported dialects.


nao1215:
All six review notes are addressed; two were defects in this branch's own code and one changed what I had written.

The commit hook was the serious one. `BEGIN; UPDATE; COMMIT; BEGIN` fired `transactionCommitted` while the second transaction was open, and the save reads the tables through that same connection, so it waited for a transaction the call was holding: the process hung rather than merely persisting uncommitted rows. Reproduced with a 25-second test timeout, whose stack is `runSeveralEffects -> transactionCommitted -> saveNow -> overwriteOriginalFiles`. The hook now runs when the string ends outside a transaction, which is the moment the single-statement path runs it, and `TestAutoSaveOnCommitWaitsForTheBatchToEnd` covers the three shapes.

The JSON argument constraints were real but not in the shape the note gave, so I measured them against SQLite rather than taking either of us on trust: `json_extract` accepts one argument and answers NULL, so it has no lower bound; `json_object` needs an even count and `json_insert` and `json_set` an odd one, and SQLite enforces all three when the query runs rather than when it is compiled. So `JSON_BUILD_OBJECT('a')` fetched a row and then failed with "json_object() requires an even number of arguments", naming a function nowhere in the caller's query -- the same class this pull request is about, one step later. The table now holds a rule rather than a list for those three, and the refusal names `JSON_BUILD_OBJECT`.

The filesystem root is fixed: an `fs.Stat` failure carries `ErrIOOperation` and the reason, rather than falling through to `ErrNoFiles`, with a regression test over an `fs.FS` that refuses every name.

The three notes on `transaction.go` that asked for a wider rework of the gate I have read and not taken: the gate is taken before the string runs whenever its first statement opens a transaction, and released by `applyEffect` through `closeTx`, so a string that opens and closes within itself holds it for the whole call rather than in pieces. That is the same contract the single-statement path has, and taking it apart per statement would mean running the statements separately, which is not what the driver does with the string.

